PR #4, opened 2018-02-21
Add Squad and Update Readme

Exonical:


Exonical:
Add 7 Days to Die

parkervcp:
Lets only make this for "squad" and not add 7 days to die at the same time.

PR #5, opened 2018-02-22
clean up some files

Aaronstar-McClure:
added the following block to the install script:

rm -r discord-bot-master/
rm Dockerfile
rm CONTRIBUTING.md
rm README.md
rm /config/.gitignore
rm CODE_OF_CONDUCT.md

NOTE licence.md was left in on purpose. 
Also, it seems the .config folder is created by the bot on startup. most likely it watches the configs file for changes and auto reloads


Aaronstar-McClure:
also, this is what ryan said about the three keys: 

RyanTheAllmighty - Today at 4:13 PM
Need them initially to generate an access token, after that it's not necessary

PR #14, opened 2018-03-01
Added Waterfall, Paper/Taco Spigot

rouing:
Added Waterfall
Added Paper Spigot
Added Taco Spigot


parkervcp:
Can we rename these to just waterfall, taco-spigot, and paper-spigot? Other than that I am mostly okay with these.

rouing:
@parkervcp Done, Done, and Done.

parkervcp:
The taco-spigot egg fails to import because the config options are all null.

If you used the "copy settings from" there may be a bug that I will be testing for now.

rouing:
Yeah, they are cascading copy settings from. Taco from Paper from regular.

rouing:
Should I isolate those? I planned on improving configs on both later but make the config changes per fork so it wouldn't be duplicate changes everywhere.

parkervcp:
Isolate them. Easiest fix there is right now.

rouing:
I will do that.

rouing:
Isolated.

rouing:
annnnnnnnd it added it to the pull request. Not sure you want that license in there or not, but if you are ok with it then I will leave it. If not, I will remove it.

PR #16, opened 2018-03-06
Waterfall support

xion87:


parkervcp:
I already have a PR for waterfall that looks good with https://github.com/parkervcp/eggs/pull/14 . So I will turn this one down for now.

PR #21, opened 2018-03-18
Dynamic Versioning for Forge

rouing:
I request that some other people test this before approving the PR.

I've changed quite a bit on this, so feedback is appreciated.


parkervcp:
I will give this a once over when I am off work tonight. I have a few other things I need to look into as well.

parkervcp:
@rouing default version should be "latest" not 1.12.2. To make sure it at least gets what is current.

rouing:
@parkervcp It seems this may be impossible to achieve because the Forge Download requires an explicit Minecraft version to be specified. The way forge download and version parsing works are that it requires a specific Minecraft version to be specified. Even though they define "latest" and "recommended", it would still need a Minecraft version tagged to that. I would have to hardcode the latest Minecraft version and that is just messy. Here is the command in question: 

`curl -sS http://files.minecraftforge.net/maven/net/minecraftforge/forge/$VERSION/forge-${MINECRAFT_VERSION}-$VERSION-universal.jar -o server.jar`

As you can see, ${MINECRAFT_VERSION} must be defined. So even if I did pickup "latest" from the bottom of the JSON, it would still need 1.12.2 defined.

PR #27, opened 2018-03-24
Added arma 3 egg

Yomanz:
I finally got time to add the Arma 3 egg :P

I'm gonna be making some different languages soon NodeJS etc.


parkervcp:
@Yomanz Can you get links for the docker container? I would like to see what we got going on here. See if we can generic the container.

Yomanz:
If i remember correctly it will work for any arma (fingers crossed) and has 1 or 2 additional dependencies due to arma using add-ons that need extra deps.

Daxorinator:
@parkervcp https://github.com/Yomanz/Containers/branches
It's using a generic container of some kind but it might be modified, not sure which one is in use.

quay.io\/daave\/containers:rvr4

Yomanz:
Image changed from alpine to Ubuntu due to a rare issue affection certain installations using this egg.

mjmfighter:
I am curious why this container uses a libnss_wrapper.so and a password.template file.  It doesn't seem necessary as I have seen other docker arma instances that don't do that

Yomanz:
It stops certain seg faults.

parkervcp:
I wanted to ask about those as well. I haven't looked into this very much yet.

mjmfighter:
What kind of seg faults does it help stop?  I have a hard time believing that arma needs a way to look up user and group permissions with a nss wrapping library.

rouing:
`ERROR: ld.so: object '/libnss_wrapper.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored. `

Yomanz:
You can safely ignore that error.

mjmfighter:
Still want to know what kind of seg faults it helps stops.  Below is another Docker Arma container that I had no trouble running alone... I just can't figure out when I port it to an egg, it breaks...

https://github.com/hasable/docker-a3-server

parkervcp:
Running that image is not running it with the panel. There are large enough differences it's not equatable.

With the new daemon though it is easier to watch for these errors.

mjmfighter:
I realize it isn't running it through the panel.  Wasn't saying it was.  Was mainly saying that it should be possible to run arma in a docker image without having some questionable workarounds to get it working

rouing:
Indeed. I am also curious as to what the workaround is for. I am going to try to setup my own instance of this and strip the workaround. If it works without issues (including mods and mission files) then I will make my own PR due to the lack of answer by @Yomanz

Yomanz:
Ok so I finally have an answer to that question, I removed everything related to the wrapper and user and group stuff. https://kibii.is-my.love/kf6tsTUQqM60iyf.png
I'm redoing the container now.
https://kibii.is-my.love/htch6mJZGX7XWOsz.png
Works fine after readding the wrapping library etc.

PR #42, opened 2018-05-17
Create egg-pix-a-r-k.json

KennyBZA:
Egg for PixARK - simple and straightforward. Allows for server password, admin password as well as ports (server port + 3 additional ports for Query, RCON and Terrain).

Docker repo can be viewed here: https://github.com/KennyBZA/pterodactyl-pixark


parkervcp:
I don't like having a image specific to a service. Can we attempt to have a generic image. I also have an images repo for this kind of thing.

KennyBZA:
Let me see what I can do about that. I'm just concerned about Wine's init which (should) be run beforehand. But I will take a look.

parkervcp:
I modified the local install I have to use the wine-source image I have in my images repo and it works. I will merge this now. I am going to be making some changes though.

PR #44, opened 2018-05-26
Fixed version parsing that was stopping installations

blastehh:


parkervcp:
I was actually working on a fix for this already.

blastehh:
Only needed to add linebreaks to the returned version list for grep to work, it was all json in a single line.

parkervcp:
I am working on a bit cleaner way as it's json and there are tools to handle this better.

parkervcp:
@blastehh I have an update for this. I will be using jq. The glibc image I use for installing is also updated for this change.

PR #48, opened 2018-06-08
factorio egg changes

blastehh:
- Changed json parsing to look for latest instead of assuming the last entry
- Removed password variable because you can't set an empty password
- Added variables to help host public servers
- Added the map creation options to use the map-gen json files
- Added the server settings launch option which is required for public servers


parkervcp:
On the `jq` you want to add `-r` instead of using sed at the end

`jq -r` gives the raw output without the `"` marks

PR #51, opened 2018-06-23
Create egg-pocketmine.json

onekintaro:
For all who need it :)


onekintaro:
Oh i put it in the wrong folger

PR #52, opened 2018-06-23
Pocketmine MP egg (Minecraft Bedrock Engine)

onekintaro:
For all who need it :) now in the correct folder :)


Tomoli75:
GG on the egg creation, someone please **accept the PR**.

r7vmc:
what about adding "--disable-ansi" to the start command? it will fix the spam
https://forums.pmmp.io/threads/is-there-a-way-to-disable-cli_set_process_title.6052/

onekintaro:
Ah this fixes the spam :) thank you i added the command to startup

parkervcp:
Fist things first I won't let you have an egg that uses a shell script to start. It's too easy to modify and have it run other things.

Second there has to be a better image for this. I have yet to look through any of it other than that.

onekintaro:
Okay i will look to fix this things :)

parkervcp:
I am trying it out as well. I am just poking around right now.

PR #54, opened 2018-07-02
RageMP Update

Exonical:
Updated egg config to remove boolean variable not compatible with config file
**Installation Update**

- Automatically download and install RageMP Bridge Plugin used for custom plugins on RageMP Server

**Container Update**

- Added NodeJS v8 Support
- Added dotNet Core Runtime 2.1


parkervcp:
I am going to say no to this as it needs a custom container.

PR #58, opened 2018-07-20
Update requirements for startup configuration.

ScottUK:
FiveM supports 32 players | Also a server name can have as many characters as possible


parkervcp:
What I saw called for 31 being the max. As for the name string that is understandable.

PR #67, opened 2018-08-14
Added Starbound

rouing:
Signed-off-by: Rouing <rouing@waterfallnetwork.net>


parkervcp:
I have made major changes to how things are organized in the repo. Can you update to match please. I would put this under source servers as it needs steamcmd to install.

Is there a way to not use user/pass auth and use a token to install?

rouing:
Currently, no there is not. I will reorg the PR

rouing:
If you do happen to find a way to bypass login, please tell me. It is the most ANNOYING thing ever.

parkervcp:
Can you add to the main README and then the local server README. Please check the warband to get a good idea.

rouing:
Ive moved away from Pterodactyl fully. This is abandoned by me. If you want you can take over this PR.

PR #68, opened 2018-08-15
Fix port used for startup - xonotic

CosmicServal:
Replace {{DEFAULT_PORT}} by {{SERVER_PORT}}


parkervcp:
`DEFAULT_PORT` should be a map for `server.build.default.port`

PR #93, opened 2018-10-16
Broken image?

Sir3lit:
The image you changed it to causes a segmentation fault, What was wrong with the original image?


parkervcp:
This is intentional.

PR #96, opened 2018-10-26
Added Holdfast: Nations At War server egg

masonr:


parkervcp:
@masonr Can I ask why the readme says this needs over 2400 ports?

masonr:
@parkervcp those are number ranges that represent acceptable port numbers for each of the three ports required. The port ranges are specified in their documentation but have not been tested to see if the server will work as expected with port numbers outside of these ranges.

I could clarify that the port number has to be picked from within the specified range by adding a "(range)" after "default" in the header row of the port table.

parkervcp:
@masonr Default to an absolute minimum. If a single port will work the lowest possible will do.

masonr:
Can do, will note the acceptable ranges in the notes section and just list single ports for the port table.

parkervcp:
@masonr We can test this out when you are available next.

masonr:
@parkervcp - should be able to sort it out one night this week if you need me.

I have a secondary dev Steam account with a Holdfast key given to me by the developers that I can share access to so that you can test installing the server files. You can either install the game locally and test connecting to the server to verify everything is good, or you can use my server banner generator (https://holdfastbanner.masonrowe.com/) that'll scrape the stats of the server once it's up and running. I've already confirmed there aren't any issues playing on the server once it's up and running, but you're more than welcome to confirm on your end as well.

There is one quirk getting the server to run. On the initial start up, it'll fail and will need restarted to start up correctly. I've chalked this up to needing the wine config files present before a successful boot can occur, but I haven't found a solution to have the server start up fine initially. Let me know if you need any help while testing.

parkervcp:
I will be closing this in favor of getting a new PR that uses the new PR template I have added to the repo.

If you still want to add this please resubmit a PR for this.

PR #112, opened 2018-11-17
Revert image back to original

Sir3lit:
Revets the image for the container back to the original one.

quay.io\/daave\/containers:rvr4


parkervcp:
So the issue I have with this is that I am using the generic image and it works fine.

parkervcp:
Closing this as I have some things to look into now after some testing was completed by AreYouScared.

PR #113, opened 2018-11-18
ARK: Fixed install script for some servers

Sporesirius:
Ark server got stucked at some servers before the changes.

Added code (for better reading):
# Some servers require these settings to run the ARK server under Linux Docker.
echo 'fs.file-max=100000' | tee -a /etc/sysctl.conf >/dev/null
sysctl -p /etc/sysctl.conf
echo '*               soft    nofile          1000000' | tee -a /etc/security/limits.conf >/dev/null
echo '*               hard    nofile          1000000' | tee -a /etc/security/limits.conf >/dev/null
echo 'session required pam_limits.so' | tee -a /etc/pam.d/common-session >/dev/null


parkervcp:
These would only affect the install image which is thrown away after the install is complete so this won't help.

PR #125, opened 2018-12-26
fix issues for ATL bot install

Sir3lit:
Allows `npm install` to run without a error


parkervcp:
@Sir3lit I believe we discussed this does not fill all the issues with the bot correct?

Sir3lit:
Correct, there are way more issues with atl then just that. I gave up trying to fix it.

parkervcp:
I am going to accept this PR as it fixes the install alone. The egg is still broken.

PR #130, opened 2019-01-02
Bastion V7 Update

Matthww:
Update eggs to support Bastion V7


parkervcp:
does v7 support both json and yaml @Matthww ??

Matthww:
He changed from json to yaml configs

parkervcp:
Because of course they did...

I'll allow it.

PR #132, opened 2019-01-04
Add OneSync, new update for FiveM which allows up to 64 slots.

ScottUK:


gta/fivem/egg-five-m.json
```diff
@@ -24,7 +24,7 @@
     },
     "variables": [
         {
-            "name": "fivem license",
+            "name": "FiveM License",
```
parkervcp: Why upper case this?

parkervcp:
@ScottUK Please explain what onesync is and why it should be added.

gta/fivem/egg-five-m.json
ScottUK: My bad, probably just my OCD I will removei t.

ScottUK:
OneSync allows for up to 64 slots from the normal 32 that FiveM offers. It is in Alpha and is more likely to cause crashes, however, it is pretty stable as I currently use it on my FiveM server.

parkervcp:
I'll say no to just updating the current egg as it could cause potential issues with installs as it's an Alpha build.

Move it to a different folder and change the name to something like FiveM-OneSync

parkervcp:
I am closing this due to lack of response.

PR #144, opened 2019-01-11
Added steam auth for starbound

NgLoader:
- You must own starbound to install it!
- It only works with the Mobile Auth App.
- When your not using any auth let it empty.
- The reason is that the Email Auth sends you the code when you're trying to login but you'll need to enter the key bevor that.


NgLoader:
Already exist

PR #173, opened 2019-03-25
Add basic configs to the Arma 3 Egg

BrainP4in:
- The installation script now downloads a default config file from the eggs repo, which improves the usability for users who are unfamiliar with Arma.

- Rename the Config files so that the names match the official documentation from Bohemia Interactive.

### All Submissions:

2. [x] Have you tested your Egg changes? -> with my git repo 😃


parkervcp:
@BrainP4in Can you make it so if a file exists you don't overwrite it.

The install right now would reset the configs is you use reinsatll to update or anything.

BrainP4in:
I have tested it today and it's fixed. Thanks for the hint.

parkervcp:
This looks good.

PR #191, opened 2019-04-29
Update ET Legacy

danny6167:
### All Submissions:

* :white_check_mark: Have you followed the guidelines in our Contributing document?
* :white_check_mark: Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?

1. :white_check_mark: Have you added an explanation of what your changes do and why you'd like us to include them?
2. :white_check_mark: Have you tested your Egg changes?

This pull request makes the following changes:
 - Correct stop command
 - Add MAP to start params and server vars so server will init
 - Work in /tmp/ to avoid leaving installers and unneeded files in the server directory
 - Slightly re-order some of the setup code
 - Add startup string matching so server can update status to running
 - Create '/mnt/server/.etlegacy' during setup needed for the servers sqlite database (server wont create it its self)
 - Correct wget command for 64 bit installs

### DIFF
--- a/etlegacy/etlegacy/egg-e-t-legacy.json
+++ b/etlegacy/etlegacy/egg-e-t-legacy.json
@@ -3,21 +3,21 @@
     "meta": {
         "version": "PTDL_v1"
     },
-    "exported_at": "2019-03-11T06:46:01-04:00",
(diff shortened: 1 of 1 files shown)


parkervcp:
I will give this a look later. Thanks for taking this on.

PR #209, opened 2019-05-26
Automatically overwrite files when reinstalling

JoshuaWalsh:
This is necessary because otherwise `unzip` asks for input, and there's no way in Pterodactyl to provide this input.

This does run the risk of destroying user customisations to configs and stuff, but the Pterodactyl UI already warns "Danger! This could overwrite server data." so I think it's acceptable.

### All Submissions:

* [✓] Have you followed the guidelines in our Contributing document?
* [✓] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?
* [✓] Did you branch your changes and PR from that branch and not from your master branch?

1. [✓] Have you added an explanation of what your changes do and why you'd like us to include them?
2. [✓] Have you tested your Egg changes?


JoshuaWalsh:
@parkervcp I've made some changes to this PR, could you please take a look at it?

winadam:
If you are doing a pr on this file, could you fix the `Assumeing` typo?
:)

JoshuaWalsh:
Just for you @winadam :)

parkervcp:
@JoshuaWalsh why not move the files instead of wiping them? at least let the user remove the files if they feel like the update worked?

JoshuaWalsh:
The rationale was to move any files (just the config directory) that are likely to have been modified by the server owner, but wipe files that just come with the modpack. If someone wanted to rollback the update they could simply reinstall whatever the previous version was. I'll change this to keep as many files as possible though.

Something I'm not happy with is the fact that the config_backup directory gets deleted if it already exists. This means if someone does 2 reinstalls in a row their original config will be lost. I'm going to fix this too.

PR #272, opened 2019-07-29
Added download link

marrobHD:


kevinsnijder:
You used the brackets wrong, you can put an x or a space in between brackets to make it a checkbox which is either checked or unchecked.
Also, why is this better?

marrobHD:
Oh sorry. I think it's better because no users need to search for a direct download link

marrobHD:
But with "latest" it wont work :(

PR #452, opened 2020-03-04
add redis egg

parkervcp:
This would add a redis egg

### All Submissions:
  * If not, why?:

### New Server Submissions:
2. [x] Does your server use a custom docker image?
    * [x] Have you tried to use a generic image?
    * [x] Did you PR the necessary changes to make it work?

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ If you are reading this it looks like you are looking to add an egg to your serv
 
 ## Database Eggs
 [MariaDB](/database/mariadb/)
-
+[Redis](/database/redis/)
 ## Game Eggs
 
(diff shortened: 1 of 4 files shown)


dimaguy:
It kinda works. Makes a server but then it gets stuck at installing till one forces install toggle(Yes I checked install.log and it says done)
Then it never fully starts redis. Gives a warning about password that shouldnt be in command(can be ignored) but then never fully starts the client
I gave it 4 GB ram and unlimited storage/CPU/swap

parkervcp:
redis is a key value store. I know it's not a database. I was just lumping it in with them.

PR #462, opened 2020-03-23
Updating FiveM egg to support OneSync and have up to 128 players

DoJoMan18:
### All Submissions:

  * If not, why?:

### New Server Submissions:

1. [x] Does your submission pass tests (server is connectable)? Yes

1. [ ] Have you added an explanation of what your changes do and why you'd like us to include them? Yes
2. [x] Have you tested your Egg changes? Yes


parkervcp:
As far as I know this is still a paid feature. Adding a variable and making it a default will either get people complaining that it isn't working on they are stuck at 32 people max when they set it higher than that.

DoJoMan18:
If they don't have the correct license to use OneSync it isn't going to work. That is also the reason that OneSync is switch off at default. However I can add a disclaimer in the discripting to inform the people that they need a specific license before OneSync and/or the users above 32 is going to work. 

If they configure OneSync and set the player limit above 32 without the correct license the server will start up without any problem.

parkervcp:
I don't want to deal with the fallout from that. I also believe OneSync is still a beta feature.

parkervcp:
this is gonna be a no for now. Wait until onesync is at least out of beta.

PR #531, opened 2020-05-21
Updated Terraria

aarroonn789:
Updated Terraria server installation egg to 1.4.0.3. If Terraria would keep all of their server files in one place this wouldn't be an issue.

### All Submissions:


iamkubi:
I've updated the egg to scrape the wiki for download links so this isn't necessary.  Will be pushed shortly, so I don't think this is needed.  Thanks though!

PR #539, opened 2020-05-26
fivem egg compatibility update to newer versions

DarKiiii:
### All Submissions:

* [ ] Have you followed the guidelines in our Contributing document?
yes
* [ ] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?
yes
  * If not, why?: no, because the onlyy changes i did are on the fivem egg

1. [ ] Have you added an explanation of what your changes do and why you'd like us to include them?
i only updated the egg to be compatible to the newer versions of fivem ( above 2417 )
yes and it works perfectly


parkervcp:
1. You filled out none of the information
2. It looks like you copy and paste my stuff into a new egg instead of editing the one I already have. Thus replacing the auth tag in the egg.
3. You opened this pull request from your master which means it will catch any other changes you make.

parkervcp:
As a side not you also give no reason as to why these changes are necessary.

PR #605, opened 2020-07-11
Update Conan Exiles

Exonical:
Update changes to Wine 5, fixes xvfb creation on startup to allow it to run due to it being broke after updates.


parkervcp:
I would rather not use a shell script here....

PR #618, opened 2020-07-16
small grammar fix

camjac251:
grammar fix


Hydra2k:
this is incorrect

camjac251:
it's is short for it is.  https://www.dictionary.com/e/its-vs-its/

Hydra2k:
it was correct originally

camjac251:
Did you happen to check out the link Hydra? I was hoping that it could offer some clarity on the difference between the two.

PR #701, opened 2020-09-22
Added ArchiSteamFarm

NgLoader:
### New Server Submissions:

1. [X] Does your submission pass tests (server is connectable)?
3. [X] Have you added the server to the main README.md?
4. [X] Have you added a unique README.md for the server you are adding?

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,9 @@ If you are reading this it looks like you are looking to add an egg to your serv
 * [TeamSpeak3](bots/teamspeak3)
   * [JTS3ServerMod](/bots/teamspeak3/jts3servermod/)
 
+* [Steam](bots/steam)
+  * [ArchiSteamFarm](/bots/steam/archisteamfarm/)
+
 ## [Database](/database/)
 ### [SQL] Databases
 * [MariaDB](/database/sql/mariadb/)
--- a/bots/steam/README.md
+++ b/bots/steam/README.md
@@ -1,0 +1,5 @@
+# Steam Bots
+
+### Some of these bots support other services but are primarily Steam bots
+
+#### ArchiSteamFarm [Homepage](https://github.com/JustArchiNET/ArchiSteamFarm) [Egg](/bots/steam/archisteamfarm/)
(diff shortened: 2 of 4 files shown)


parkervcp:
Why does this need a custom image?

https://github.com/parkervcp/images/tree/debian/dotnet

NgLoader:
ArchiSteamFarm run only on x64 arch and you need to set the DOTNET_BUNDLE_EXTRACT_BASE_DIR to another directory because of permission issues.
When you have any idea how I can change that, let me know that please.

parkervcp:
You can set the ENV vars in the egg as variables.

NgLoader:
Okay, I test it tomorrow

NgLoader:
> You can set the ENV vars in the egg as variables.
Okay, thanks

I have changed the image to https://github.com/parkervcp/images/tree/debian/dotnet and tested it.

parkervcp:
Start scripts are usually user modifiable. can we get rid of that?

NgLoader:
Like that?
And yes we can change it and it works for me.
But the issue https://github.com/JustArchiNET/ArchiSteamFarm/issues/1812 is still there I think.

NgLoader:
Okay, looks like the problem is fixed. so we can run it on Docker with no problem.

tiweb442:
i try this pull on my panel and its working fine whit no bug

parkervcp:
So while it is a working egg I don't know how steam feels about this tool and am not really comfortable with it.

parkervcp:
Going to close this out as I don't want to clash with valve with this if it ever becomes an issue.

NgLoader:
I found this comment on steam by an administator

> It's entirely impossible you got banned over a single automated trade. ASF is a grey area, but Archi is in direct contact with Valve, and ensures that nothing violates ToS at all times, and if it does, it would be disabled prior to anyone actually being punished for it.
The only reports I've seen of people getting banned is for such stupid things as using !loot asf with 100+ bots.
[Reference](https://steamcommunity.com/groups/archiasf/discussions/6/1694922526905360875/)

parkervcp:
as stated - if it ever becomes an issue.

PR #787, opened 2020-11-12
Removed, Useless

Lycol50:
you can't connect to it if you change the Server IP.

### All Submissions:

  * If not, why?:

### New Server Submissions:

### DIFF
--- a/among_us/impostor_server/egg-among-us--impostor-server.json
+++ b/among_us/impostor_server/egg-among-us--impostor-server.json
@@ -10,7 +10,7 @@
     "image": "quay.io\/parkervcp\/pterodactyl-images:debian_dotnet",
     "startup": ".\/Impostor.Server",
     "config": {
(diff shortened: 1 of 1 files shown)


parkervcp:
1. How often are you changing the IP?
2. The public ip is outside the container and this guarantees that the right ip gets advertised.

Lycol50:
1. Not much
2. but the problem is if the port changed and the IP of that node (via Network Allocation on panel) is set to 0.0.0.0, it will just not work, because 0.0.0.0 is used by the listener.
- Changing IP on the config file, will not just work, it will just revert it back based on the config script of the egg every server restart.

parkervcp:
The ip is set from the allocations which is how IP changes should be made,

Lycol50:
yeah, but to be able to broadcast the server, the PublicIP should be set to 127.0.0.1, instead of 0.0.0.0

gOOvER:
That was on my first commit 127.0.0.1, but someone told me to make public ip changeble

PR #861, opened 2020-12-13
upgrade to postgres 13.1

nicolas-janzen:


parkervcp:
1. You didn't fill out the template.
2. All this does is fix the error about the db being built on the old version of postgres while it's still not usable.

nicolas-janzen:
I am sorry I did not fill out the template. Have you double checked that?
The egg didn't work with the old version in our setup so we just changed the version and got Postgres to work.
I thought it would be appropriate to share this fix with you.
Have a nice day!

parkervcp:
The egg didn't work before that either.

I am working on making it better.

PR #881, opened 2020-12-25
Add support for sky factory 4.

Zarklord:
adds support for the modpack Sky Factory 4, which doesn't work under normal install methods (failing to install when calling "mv FTBServer.jar server.jar" before)

### All Submissions:


parkervcp:
I don't like this approach of making a specific change for a single pack. That will make this huge.

PR #882, opened 2020-12-25
pause_when_empty defaults to true.

Zarklord:
pause_when_empty should by default be true, it is the default and intended behavior for 99% of servers, along with that, most players(myself included), expect that default configs will match the standard default of the game, and this caught me off guard to see the game was still running.

### All Submissions:

### DIFF
--- a/steamcmd_servers/dont_starve/server.cluster.ini
+++ b/steamcmd_servers/dont_starve/server.cluster.ini
@@ -2,7 +2,7 @@
 game_mode = survival
 max_players = 10
 pvp = false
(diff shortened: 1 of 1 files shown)


parkervcp:
1. That config is one I got from my server. It's the default setting.

2. You gotta fill out all the parts for the templates.

parkervcp:
I'm meeting this because it's not a bad ideas anyways.

Zarklord:
well, you accepted it, so it doesn't matter, but in general, I would only expect to need to make config changes if it goes against the game defaults.

PR #901, opened 2021-01-02
Add multi-image selection to a bunch of eggs

TekExplorer:
### All Submissions:

  * If not, why?:

Simply enough, adding multiple images to a bunch of eggs that can benefit from it while leaving `image` alone for older versions of the panel.

I'm not sure what image would be used as the default, though I think there really should be a way to set a default and alias for each image, so if I need to re-order something lmk.

Also switched some eggs from using parker's java 11 image to the official pterodactyl one, cause why not.

Not sure if I missed anything, and I have not tested them, as I don't have the time, though it looks right.


parkervcp:
While I support this idea I want them to be separate PR's for each egg type. MC vs bot eggs. @TekExplorer

TekExplorer:
Alright, might get around to that when I get home. Should I also add some of your other jdk versions to the mc ones?

TekExplorer:
Closed in favor of #910 and #911

PR #908, opened 2021-01-04
Update egg-velocity.json - propagate IP binding into velocity.toml on startup

ens-gijs:
Respect IP binding when modifying velocity.toml on startup

### All Submissions:

  * If not, why?:


TASelwyn:
Sorry, what's wrong with the server instance binding to 0.0.0.0? Due to how Pterodactyl works, it only allows IP/ports that are allocated to the server directly, which would be 123.456.789.0 or whatever..... I don't see an actual change for this except to prevent local connections from the pterodactyl wings network (aka another server on the same node)

parkervcp:
This will fail unless you run your node in host mode.

PR #935, opened 2021-01-21
phantombot.tv switched to github.io

ExXxO:
Changing Link in .json due changing from phantombot.tv to github.io


eramsorgr (approved):
Domain change is indeed correct

PR #965, opened 2021-02-05
Add Valheim

acl1704:
### All Submissions:

* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?
* [x] Did you branch your changes and PR from that branch and not from your master branch?
  * If not, why?:

### New Server Submissions:

1. [ ] Does your submission pass tests (server is connectable)?
2. [ ] Does your server use a custom docker image?
    * [ ] Have you tried to use a generic image?
    * [ ] Did you PR the necessary changes to make it work?
3. [x] Have you added the server to the main README.md?
4. [x] Have you added a unique README.md for the server you are adding?

Draft, requires testing as I do not own the game.

Closes #959

PR #968, opened 2021-02-06
Add Valheim

### LINKED ISSUE #959: Egg Request: Valheim
Please fill out the information bellow and remove from the line up
Please understand how Pterodactyl works when you are requesting an egg. (ie. docker-compose doesn't work for a pterodactyl server)
---------------

Service: Valheim dedicated server

Does this expand an already existing service: Y, it's SteamCMD, linux compatible

Link to game: https://store.steampowered.com/app/892970/Valheim/

Links for server downloads:  https://steamdb.info/app/896660/

Links for install steps/docs: https://steamcommunity.com/sharedfiles/filedetails/?id=2382282960 (usual SteamCMD setup really)

Magi1053:
I started off with the egg from #965 but it was semi non functional, so I rewrote the startup script and made some other changes such as actually outputting the server log and using the native server shutdown trigger

Closes #959

### All Submissions:

### New Server Submissions:

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,7 @@ If you are reading this it looks like you are looking to add an egg to your serv
 * [The Forest](/steamcmd_servers/the_forest)
 * [Team Fortress 2 Classic](/steamcmd_servers/team_fortress_2_classic)
 * [Unturned](/steamcmd_servers/unturned)
+* [Valheim] (/steamcmd_servers/valheim)
 
 [Teeworlds](/teeworlds)
 * [Teeworlds](/teeworlds/teeworlds)
--- a/steamcmd_servers/valheim/README.md
+++ b/steamcmd_servers/valheim/README.md
@@ -1,0 +1,11 @@
+# Valheim Server
+A brutal exploration and survival game for 1-10 players, set in a procedurally-generated purgatory inspired by viking culture. Battle, build, and conquer your way to a saga worthy of Odin’s patronage!
+
+https://store.steampowered.com/app/892970/Valheim/
+
+## Server Ports
+
+| Port  | default |
+|-------|---------|
+| Game  | 2456   |
+| Query | 2457   |
--- a/steamcmd_servers/valheim/egg-valheim.json
+++ b/steamcmd_servers/valheim/egg-valheim.json
@@ -4,15 +4,13 @@
         "version": "PTDL_v1",
         "update_url": null
     },
-    "exported_at": "2021-02-07T12:16:16-06:00",
+    "exported_at": "2021-02-07T19:36:46-06:00",
     "name": "Valheim",
     "author": "magi1053@outlook.com",
     "description": "A brutal exploration and survival game for 1-10 players, set in a procedurally-generated purgatory inspired by viking culture.",
     "features": null,
-    "images": [
-        "quay.io\/parkervcp\/pterodactyl-images:debian_source"
-    ],
-    "startup": "{{INIT}}; .\/valheim_server.x86_64  -nographics -batchmode -name \"{{SERVER_NAME}}\" -port {{SERVER_PORT}} -world \"{{WORLD}}\" -password \"{{PASSWORD}}\" -public 1 |  sed -uE \"{{FILTER}}\" & trap \"echo 1 > server_exit.drp && wait\" 15; wait $!",
+    "image": "quay.io\/parkervcp\/pterodactyl-images:debian_source",
+    "startup": ".\/valheim_server.x86_64  -nographics -batchmode -name \"{{SERVER_NAME}}\" -port {{SERVER_PORT}} -world \"{{WORLD}}\" -password \"{{PASSWORD}}\" -public 1 |  sed -uE \"{{FILTER}}\" & trap \"echo 1 > server_exit.drp && wait\" 15; wait $!",
     "config": {
         "files": "{}",
         "startup": "{\r\n    \"done\": \"DungeonDB Start\"\r\n}",
@@ -38,7 +36,7 @@
         },
         {
             "name": "Server Password",
-            "description": "Server password.\"",
+            "description": "Server password.",
             "env_variable": "PASSWORD",
             "default_value": "secret",
             "user_viewable": true,
@@ -64,10 +62,10 @@
             "rules": "nullable|numeric"
         },
         {
-            "name": "Before Server Start",
-            "description": "Helper script to manage the server instance",
-            "env_variable": "INIT",
-            "default_value": "export templdpath=$LD_LIBRARY_PATH; export LD_LIBRARY_PATH=.\/linux64:$LD_LIBRARY_PATH; export SteamAppId=892970",
+            "name": "LD Library Path",
+            "description": "Required to load server libraries.",
+            "env_variable": "LD_LIBRARY_PATH",
+            "default_value": ".\/linux64",
             "user_viewable": false,
             "user_editable": false,
             "rules": "required|string"
--- a/steamcmd_servers/valheim/egg-valheim.json
+++ b/steamcmd_servers/valheim/egg-valheim.json
@@ -4,7 +4,7 @@
         "version": "PTDL_v1",
         "update_url": null
     },
-    "exported_at": "2021-02-06T12:50:53-06:00",
+    "exported_at": "2021-02-06T19:50:18-06:00",
     "name": "Valheim",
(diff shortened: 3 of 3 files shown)


Software-Noob:
Remove file_denylist or otherwise the egg importing will not work for the latest panel versions until the next release when that issue is resolved.

Software-Noob:
You have validate flag in your install, all eggs by default do not perform any destructive actions on reinstall. Your startup exports of the LD path could just be a variable. I do not see any reason why are you exporting auto_update or APP_ID?

Magi1053:
@Software-Noob What is the reason for putting the startup exports into a variable instead of leaving them them as is?

Software-Noob:
> @Software-Noob What is the reason for putting the startup exports into a variable instead of leaving them them as is?

It's pointless clutter in the startup for no reason, your three export lines for LD LIB PATH could just be one hidden variable.

Magi1053:
I'm getting syntax errors putting the multiple exports into 1 variable, could you provide me with an example?

Magi1053:
So the server spits out a ton of useless debug info that I was piping to sed to filter out, but I can't seem to get that to work when it's inside of the variable

Magi1053:
@Software-Noob Take a look at that and lmk what you think

Magi1053:
The shutdown trigger is not functional, working on a fix.

Software-Noob:
Your INIT variable for LD LIB PATH could just be set to ./linux64 and renamed to LD_LIBRARY_PATH as done in other eggs such as [here](https://github.com/parkervcp/eggs/blob/77ea6c770a428f130cc4c7637801084be511c8b4/steamcmd_servers/avorion/egg-avorion.json#L135) and [here](https://github.com/parkervcp/eggs/blob/77ea6c770a428f130cc4c7637801084be511c8b4/steamcmd_servers/7_days_to_die/egg-7-days-to-die.json#L63 ). You also don't have to include it in the startup for exporting.

Other than that LGTM except for that double quote in the server password description. The server starts, can be joined, and stops as it should.

gOOvER:
I see no need to use any external Image. As i see you can also use Debain:source

gOOvER:
Some Errors:

`./steam/steamcmd.sh: line 37: /home/container/steam/linux32/steamcmd: No such file or directory
:/home/container$ ./start_headless_server.sh -nographics -batchmode -name TestServer -port 2456 -public 1 echo Server started
/entrypoint.sh: line 13: ./start_headless_server.sh: No such file or directory`

Headless dont existst after download

gOOvER:
Next thing is; You know, that you need Mono? Your Egg will never work. You Image is a normal Ubuntu 20.04 Source Image.

gOOvER (changes requested):
Some Errors:

./steam/steamcmd.sh: line 37: /home/container/steam/linux32/steamcmd: No such file or directory :/home/container$ ./start_headless_server.sh -nographics -batchmode -name TestServer -port 2456 -public 1 echo Server started /entrypoint.sh: line 13: ./start_headless_server.sh: No such file or directory

There is no start_headless_server.sh after download

Only a start_server.sh. Pls Update, because i don't found such an Linus Server in any Guides

In this state, your egg is complete broken

gOOvER (changes requested):
Next thing is; You know, that you need Mono? Your Egg will never work. You Image is a normal Ubuntu 20.04 Source Image.

Magi1053:
> There is no start_headless_server.sh after download Only a start_server.sh.

start_server.sh is part of the server download and it the egg does not use it at all. 

>./start_headless_server.sh: No such file or directory

I have no idea why you are starting your server with that file, it is not part of the server download and my egg does not create it or use it at all. `start_headless_server.bat` is a file that is in the windows version of the server, but this egg uses the linux version. 

The egg launches with the native linux binary `./valheim_server.x86_64  -nographics -batchmode...`. I'm not sure if mono is required in the image or not, but the server definitely does start and is joinable and @Software-Noob [confirms it](https://github.com/parkervcp/eggs/pull/968#issuecomment-774762575).

arriej:
Container is running fine for me, unfortunately not joinable, port stays closed.

xardoniak:
Container _appears_ to be running, but is not joinable.
Hopefully these screenshots are useful!
image
image
image
image

xardoniak:
Same as above - server appears to be working, but is not joinable. Screenshots,
image
image
image

Magi1053:
@arriej @xardoniak When using the steam server browser you need to use the query port 2457 not the game server port.

xardoniak:
@Magi1053 Beautiful! Good old user error!
image
Working in game now, too. Not sure what the difference was before and now as it wasn't showing up earlier today. Maybe the server takes some time to auth against the master list??
image

Magi1053:
Yeah their in game server list is kinda poo imo

Magi1053:
Should I maybe swap around the game server and query ports so that people don't get confused?

xardoniak:
Might be a good idea!

I know this isn't the best place for this but unsure how to DM you! What's your thoughts on hosting 2 instances of the server on the same IP? I'm wanting to set up a public server AND a private sever. Is it as simple as just spinning up another instance, then changing the 3 IPs to something else? eg 2458, 2459, 2460?

Magi1053:
I thought about swapping the game server and server query ports, but looking at other example eggs such as Ark the primary port is the game server port and the server query port is a variable. So based on that I think it would be good to keep it the way it is.

arriej:
I try to ping the port directly on the machine, after it not responding in the server browser. And when I check 2456, 2457 or 2458 they are staying closed. This is the only egg that has this issue.

Software-Noob:
> I try to ping the port directly on the machine, after it not responding in the server browser. And when I check 2456, 2457 or 2458 they are staying closed. This is the only egg that has this issue.

The egg does not have any control over ports, if you allocated the ports then Pterodactyl should open them unless you have a firewall blocking it.

arriej:
Then I'm not sure where the issue is then. I've tried to test this other egg "https://github.com/RestoreMonarchy/pterodactyl-Valheim" and that one opens the ports fine but the server doesn't work for me.

grimesy88:
I have the egg running without issue. I had an issue with ports but what I had to do was ensure I give the server two allocations. That way the egg will use the first port and second port for game and query. When connecting to the server via Steam > View > Servers you need to use the second port to connect. So from an egg point of view it all seems good. Now as for the game itself and its severe desync problems, that's another story!

AvonGenesis:
How many ports does this egg officially need? I ran a netstat in the container and only saw it listening on one port (game server port +1)

Spent about an hour troubleshooting to find that the game was accepting connections on a different port from what was configured in the Pterodactyl interface.

Software-Noob:
> How many ports does this egg officially need? I ran a netstat in the container and only saw it listening on one port (game server port +1)
> 
> Spent about an hour troubleshooting to find that the game was accepting connections on a different port from what was configured in the Pterodactyl interface.

Two ports as listed in readme: Game port and query port which is game port +1 with query port being the port used for connections.

Butain:
@Magi1053  This is working great for me so far, thanks for working on it!

xardoniak:
@Magi1053 I've been using this successfully for a little under a week now.
Some issues I've come across,

1. Server commands don't seem work: https://www.yekbot.com/valheim-admin-commands/
The console doesn't output anything for info or help, for example
2. We seem to lose some world data (maybe 30 mins?) on reboot (eg building progress, chest contents etc)

Magi1053:
@xardoniak thank you for your concerns. First off, the dedicated server admin console does not respond to any input from the user. All admin commands in the example you listed are used in game only, that guide even tells you how to do it. Second regarding the server stopping gracefully, we are using the SIGINT signal to gracefully shutdown the server. Please make sure your egg and variables are up to date, the shutdown command is functional with the latest server version.

xardoniak:
> @xardoniak thank you for your concerns. First off, the dedicated server admin console does not respond to any input from the user. All admin commands in the example you listed are used in game only, that guide even tells you how to do it. Second regarding the server stopping gracefully, we are using the SIGINT signal to gracefully shutdown the server. Please make sure your egg and variables are up to date, the shutdown command is functional with the latest server version.

Yeah I've come across a few servers that are like this recently, which seems a little silly. I guess I've been spoiled by the Minecraft server software.

I thought I had the latest version of the egg, but it turns out I didn't! I haven't had the opportunity to test shutting down a server and seeing if data stays but I assume its fine.

PR #970, opened 2021-02-08
Fix Eco start done config

Magi1053:
The server is outputting `[Server Initialization]...` when it first starts and not when it finishes starting. Fix server startup done config by adding `Finished in`.

### All Submissions:

### DIFF
--- a/steamcmd_servers/eco/egg-eco.json
+++ b/steamcmd_servers/eco/egg-eco.json
@@ -3,7 +3,7 @@
     "meta": {
         "version": "PTDL_v1"
     },
-    "exported_at": "2021-02-07T19:56:46-06:00",
+    "exported_at": "2021-02-08T10:56:34-06:00",
     "name": "Eco",
     "author": "info@goover.de",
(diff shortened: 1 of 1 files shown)


gOOvER:
There are more languages than Englisch. When an other language is used, this will break the Server started

acl1704:
There are already plenty of eggs that rely on English output. If this makes the status change work as expected, then I think that's fine. Right now the functionality isn't correct.

gOOvER:
Sry, but the Egg wirk as it should. This change is not needed and break a lot

gOOvER:
What you can use:

Webserver now listening

gOOvER:
> There are already plenty of eggs that rely on English output. If this makes the status change work as expected, then I think that's fine. Right now the functionality isn't correct.

It is and only one will have this; so he can change it to the line he want and no need to break it for others.

And pls show me the other eggs ;)

ECO Server is other than other Server. THe whole output is translated, when u change the Language

gOOvER (changes requested):
This break the Startup fopr other languages than english

You can use: `Webserver now listening` insteed

Magi1053:
It's too bad that it doesn't support regex

Software-Noob:
> It's too bad that it doesn't support regex

It will match the string, doesn't have to be a full exact string. Arrays can also be used to configure multiple different strings.

PR #971, opened 2021-02-08
Add client port parameter to Garry's Mod server

Boy132:
This PR adds a client port variable to the garry's mod egg. This is usefull if you want to have multiple servers with the same ip but different ports.

### All Submissions:

* [X] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?


danny6167:
Are you running host mode networking? This shouldn't be necessary at all otherwise.

Boy132:
> Are you running host mode networking? This shouldn't be necessary at all otherwise. Even then I'm doubtful.

In the past I had problems with this. But seems like this is no longer necessary indeed.

PR #985, opened 2021-02-17
Update intaller script

adrijanb:


Software-Noob:
What's the reason behind this request? There's no difference between ubuntu 18.04 or 20.04 in the install container, also all new eggs move towards using debian:buster-slim for installations

adrijanb:
In the old installation process the command "apt -y update" was used, unfortunately the command did not work and leads to errors because it cannot update the packages. So I changed the command in the installation command to "apt update".

The changed Ubuntu version only serves to ensure that the container uses the latest version.

Software-Noob:
Oh, I just noticed that it's using 18.10 cosmic release and not 18.04, which is pretty stupid.

 I'd recommend changing to `debian:buster-slim` as that's the lightweight image that all new containers are moving towards

parkervcp:
I agree with converting to `debian:buster-slim` from ubuntu

parkervcp:
@adrijanb Please see my prior comment about using the debian image instead of ubuntu.

parkervcp:
As the PR template wasn't filled in I am closing this.

PR #1016, opened 2021-03-02
Fix FTB Modpack Installation Script

NyCodeGHG:
### All Submissions:

FTB updated their server installation script and that broke the way the egg was using it. There's now an option '--auto' for scripts to use the default options instead of having to answer the questions via a script.


UndarkAido:
The works for me and would close #1018

PR #1020, opened 2021-03-05
Fixed the display of the ip address on the server list.

MinerPL:
Adding the startup parameter `+ set sv_listingIPOverride {{SERVER_IP}}` with each query to the fivem servers, it sends them with the IP address given in the parameter so that the server displays with the correct IP address.

CMD - Machine Address
Left Screen - Server
Right Screen - KeyMaster and heartbeat IP Addess

image

### All Submissions:

  * If not, why?:


parkervcp:
This will cause issues with people that are hosting from their house behind a router.

MinerPL:
In what sense will it cause problems? Adding this start parameter only corrects the address in the fivem database.

parkervcp:
If you host at home behind a firewall and the server ip is `192.168.1.12` then the fivem list will still be wrong

PR #1052, opened 2021-03-26
CREWLINK - FIX - Set ADRESS to require

lucas-luchack:
Crewlink **requires** the `ADRESS` value to be provided


camden-git (approved):
I approve.

PR #1090, opened 2021-04-11
Add "-disableluarefresh" to gmod startup

Boy132:
### All Submissions:

### DIFF
--- a/stock-eggs/source-engine/egg-garrys-mod.json
+++ b/stock-eggs/source-engine/egg-garrys-mod.json
@@ -8,7 +8,7 @@
     "author": "support@pterodactyl.io",
     "description": "Garrys Mod, is a sandbox physics game created by Garry Newman, and developed by his company, Facepunch Studios.",
     "image": "quay.io\/pterodactyl\/core:source",
-    "startup": ".\/srcds_run -game garrysmod -console -port {{SERVER_PORT}} +ip 0.0.0.0 +host_workshop_collection {{WORKSHOP_ID}} +map {{SRCDS_MAP}} +gamemode {{GAMEMODE}} -strictportbind -norestart +sv_setsteamaccount {{STEAM_ACC}} +maxplayers {{MAX_PLAYERS}}  -tickrate {{TICKRATE}}",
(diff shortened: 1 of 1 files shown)


parkervcp:
@Boy132 can you explain why this would be needed? If it's useful to you but it breaks others server than I will say no.

Boy132:
> @Boy132 can you explain why this would be needed? If it's useful to you but it breaks others server than I will say no.

This is pretty much a default setting for every server to prevent reloading lua when uploading stuff.
Also not disabling it will most likely break stuff - especially when using darkrp as gamemode.

parkervcp:
Well that at least makes sense, but this should a toggle and not a default that can't be changed.

stock-eggs/source-engine/egg-garrys-mod.json
```diff
@@ -8,7 +8,7 @@
     "author": "support@pterodactyl.io",
     "description": "Garrys Mod, is a sandbox physics game created by Garry Newman, and developed by his company, Facepunch Studios.",
     "image": "quay.io\/pterodactyl\/core:source",
-    "startup": ".\/srcds_run -game garrysmod -console -port {{SERVER_PORT}} +ip 0.0.0.0 +host_workshop_collection {{WORKSHOP_ID}} +map {{SRCDS_MAP}} +gamemode {{GAMEMODE}} -strictportbind -norestart +sv_setsteamaccount {{STEAM_ACC}} +maxplayers {{MAX_PLAYERS}}  -tickrate {{TICKRATE}}",
+    "startup": ".\/srcds_run -game garrysmod -console -port {{SERVER_PORT}} +ip 0.0.0.0 +host_workshop_collection {{WORKSHOP_ID}} +map {{SRCDS_MAP}} +gamemode {{GAMEMODE}} -strictportbind -norestart +sv_setsteamaccount {{STEAM_ACC}} +maxplayers {{MAX_PLAYERS}}  -tickrate {{TICKRATE}} -disableluarefresh",
```
Software-Noob, line 11: You should create a new egg variable to enable/disable this option, for example a Boolean or a string comparison.

`$( [ "$DISABLE_LUA" == "0" ] || printf %s '-disableluarefresh' )` as example

PR #1093, opened 2021-04-11
Fix Tekkit egg

ThybeVB:
### All Submissions:

* [X] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?
  * If not, why?: I wasn't aware of needing to go on another branch, if necesarry I could always do so if requested.

The previous egg was simply not working, I'm not quite why sure why it didn't because I didn't test every bit of its steps, all I did was change it to be more up to date using examples from other eggs, and everything seems to work fine now.


parkervcp:
This changes nothing about the script but changes to use wget instead of curl in debian. After looking at the script, all that is needed it so add an `L` to the curl command

PR #1124, opened 2021-04-27
Remove useless rm call in Teamspeak3 install script

peter-r-g:
The tar file is piped straight to the tar command so no file is created that needs to be removed. Pretty useless PR for ptero but for people using Wisp will encounter a failed install everytime due to exit code 1 from the rm command trying to delete a file that doesn't exist.

### All Submissions:


Software-Noob (approved):

Software-Noob:
The stock-eggs are just copies from the [official stock eggs](https://github.com/pterodactyl/panel/blob/develop/database/Seeders/eggs/voice-servers/egg-teamspeak3-server.json) - where they should instead be updated

PR #1139, opened 2021-05-08
Added egg for Krypton

BomBardyGamer:
### All Submissions:

### New Server Submissions:

2. [] Does your server use a custom docker image?

This PR adds an egg for some new server software I'm working on called Krypton. Just to be clear, this isn't a new Spigot fork or anything, this is something completely separate. You can find the repository for the project [here](https://github.com/KryptonMC/Krypton) if you would like to know more.

The script is a bit different from what most of the other scripts seem to contain. This is due to the artifacts for the server on Jenkins always being named "Krypton-VERSION.jar", which requires us to retrieve the name of the JAR from the REST API, and then download that JAR. There might be improvements that can be made to this script, especially with the downloading and querying the REST API part, and if you think you have a better script, feel more than welcome to suggest it.


Software-Noob:
Why are you not parsing the config to overwrite the values such as port?  This will cause users to report issues with their server, as it's the expected behavior. You're providing a conf file in your PR but not using it for anything in the egg.

The egg currently does not work because the startup variable for JAR is invalid (should be `{{SERVER_JAR}}` not `{{SERVER_JARFILE}}` ). It should be made clear in the readme or the egg that users must create and provide their own world; otherwise, their server will fail to start.

BomBardyGamer:
> Why are you not parsing the config to overwrite the values such as port? This will cause users to report issues with their server, as it's the expected behavior. You're providing a conf file in your PR but not using it for anything in the egg.
> 
> The egg currently does not work because the startup variable for JAR is invalid (should be `{{SERVER_JAR}}` not `{{SERVER_JARFILE}}` ). It should be made clear in the readme or the egg that users must create and provide their own world; otherwise, their server will fail to start.

First point: I tried to use config parsing, however I was told it was very buggy and it didn't seem to work in my testing. It would be greatly appreciated if you could help me out here.

Second: I forgot to change that, my bad. I'll change that now. And I'll add that info to the README.

Software-Noob:
Hi, please add the egg to the missing readme file. It's missing from the `minecraft/java/README.md`

BomBardyGamer:
Done

Software-Noob:
There's one in `minecraft/java/README.md` that is missing

BomBardyGamer:
Done

Software-Noob:
Hey, could you please correct the conflicts as the repository structure was reorganized? After that, LGTM.

BomBardyGamer:
Sorry about the delay, been rather busy, I just solved everything, so this should be fine to merge now, if you're happy with it.

Software-Noob:
Sorry, I missed your comment in my notifications. It looks like you removed the important bit from readme:

> It should be made clear in the readme or the egg that users must create and provide their own world; otherwise, their server will fail to start.

BomBardyGamer:
Done

PR #1173, opened 2021-06-08
Neosvr beta headless client/server

Synkstar:
### All Submissions:

* [X ] Have you followed the guidelines in our Contributing document?
* [X ] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?
  * If not, why?:

### New Server Submissions:

1. [ X ] Does your submission pass tests (server is connectable)?
2. [ X ] Does your server use a custom docker image?
    * [ X ] Have you tried to use a generic image?
    * [ X ] Did you PR the necessary changes to make it work?
3. [ X ] Have you added the server to the main README.md?
4. [ X ] Have you added a unique README.md for the server you are adding?

### DIFF
--- a/steamcmd_servers/neosvr/egg-neos-vr-headless-client.json
+++ b/steamcmd_servers/neosvr/egg-neos-vr-headless-client.json
@@ -1,0 +1,59 @@
+{
+    "_comment": "Egg for neos vr headless client/server",
+    "meta": {
+        "version": "PTDL_v1",
+        "update_url": null
+    },
+    "exported_at": "2021-06-08T09:15:41-04:00",
+    "name": "NeosVR Headless Client",
+    "author": "synk@nasulex.net",
+    "description": "Headless client for NeosVR.",
+    "features": null,
+    "images": [
+        "nasulex\/neos:latest"
+    ],
+    "file_denylist": [],
+    "startup": ".\/RunHeadlessMono.sh",
+    "config": {
+        "files": "{}",
+        "startup": "{\r\n    \"done\": \"World running...\",\r\n    \"userInteraction\": []\r\n}",
+        "logs": "{}",
+        "stop": "^C"
+    },
+    "scripts": {
+        "installation": {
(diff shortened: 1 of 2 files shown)


steamcmd_servers/neosvr/egg-neos-vr-headless-client.json
```diff
@@ -0,0 +1,59 @@
...
+        "version": "PTDL_v1",
+        "update_url": null
+    },
+    "exported_at": "2021-06-08T09:15:41-04:00",
+    "name": "NeosVR Headless Client",
+    "author": "synk@nasulex.net",
+    "description": "Headless client for NeosVR.",
+    "features": null,
+    "images": [
+        "nasulex\/neos:latest"
```
Software-Noob, line 13: Why does this require a custom image? Are you sure the existing images are not enough? If so, you should PR the changes to the[ image repository.](https://github.com/parkervcp/images)

Software-Noob (changes requested):

danny6167:
This egg requires host mode networking and won't work with most peoples setups.
Calling a .sh script to start instead of having the startup command in the egg is also typically a no-no here.

Synkstar:
Can you make it run steamcmd from the start command too ?

steamcmd_servers/neosvr/egg-neos-vr-headless-client.json
Synkstar: It seems like I would need my own image because it crashes on the mono core one so I guess ill do that then + the entry point doesnt run steamcmd so I can have it do that too

Synkstar:
> This egg requires host mode networking and won't work with most peoples setups.
> Calling a .sh script to start instead of having the startup command in the egg is also typically a no-no here.

The .sh script comes with neos

elitetheespeon:
Is this just missing the docker image? I have my own and can provide the steps needed to make it work (the missing libs in the image)

Synkstar:
Idk whats going to really happen with this. I was a patreon but im not anymore. Got hosting bills to pay and I don't really play this much because no one else really does. Really in my opinion before having this egg on pterodactyl it should be publicly available and not locked behind a patreon. Right now its really hacky and you need another steam account with steam guard disabled.

parkervcp:
@Synkstar Think you can make your changes to the mono-6 yolk instead of the images change you said you made? We can work on importing this then.

PR #1177, opened 2021-06-09
patch: steamapps hotfix for steamcmd

Software-Noob:
### All Submissions:

  * If not, why?:

Only tested eggs for which I modified the old steam install script to use the correct path (/steamcmd vs /steam) or eggs for which the install image was updated to `debian:buster-slim`. Everything should work as-is since only mkdir -p was added but another set of eyes wouldn't hurt.


Software-Noob:
The hotfix is no longer relevant as steamcmd has patched this. Let's hope it won't occur again.

PR #1180, opened 2021-06-11
Add Java 16 to build Spigot 1.17

xcgc:
### All Submissions:

  * If not, why?:

This pull request adds Java 16, which is required for Spigot 1.17


Software-Noob:
It would be a better solution to compare if the version variable is latest or higher/equal to 1.17 before downloading files.

This makes it more efficient without wasting space and time. It most likely can be a package from debian.

xcgc:
> It would be a better solution to move the java downloads into the version comparisons where you export the paths. It saves time and space when both are not required.
> 
> Also, you should future-proof the comparisons instead of hard-coding them. Currently, you check for `latest` or `1.17*`. It should check for the latest or greater than 1.17, such as bc/awk or multiple string comparisons (1.17* || 1.18*). Otherwise, this will break when 1.18 is released.

Done

minecraft/java/spigot/egg-spigot.json
```diff
@@ -22,9 +26,9 @@
...
+            "container": "debian:buster-slim",
```
NgLoader, line 30: just use openjdk:16-jdk-alpine and remove from script the apk install javajdk8

Software-Noob: Java 8 is required for older spigot versions.

NgLoader: Okay, I tested it
Buildtools need java 8 to build the older versions

NgLoader: So, shut we add a new agg for newer spigot versions?
Or we need the check with version we are using and downloading the correct java version.

NgLoader:
Thats looks better

NgLoader (approved):

Software-Noob:
Thank you for your contribution!

PR #1191, opened 2021-06-20
Updated stop command

DAThosting:
Added "graceful" to stop command.
Without server won't stop.


parkervcp:
Please fill out the template.

PR #1211, opened 2021-06-28
Working ClassiCube egg.

koutsie:
### All Submissions:

### New Server Submissions:

This is a replacement for https://github.com/parkervcp/eggs/pull/1097


Software-Noob:
What is that echo install script spam? Do not use Master branch nor put your host ads into the egg.

koutsie:
Let me get rid of that.

Software-Noob:
Thanks. A few more things:

I'd recommend implementing a [GitHub release grabber](https://github.com/parkervcp/eggs/blob/master/scripts/github_release_grabber.sh) to allow users to specify the version. You don't need mono-runtime in the install script itself as you are only pulling files, nor .dll files need to be executable (it doesn't affect anything to have it but is unnecessary). Finally, deleting the .zip file at the end of the install will save space.

You are overwriting the config motd and server-name on each start but don't provide any variables for users to change the values. This will most likely lead to frustration and unnecessary support requests. 

Readme listing should be in alphabetical order, putting this after Call of Duty 4X and not Minetest.

Once those are solved, LGTM 👍

PR #1236, opened 2021-07-12
This change allows users to specify any 1.17.* version for building

Greaper88:
Have verified the following build version options on Ubuntu 20, using Pterodactyl 1.4.1:
"latest": installs openjdk 16, and runs buildtools with "--ver latest" which is a valid option (currently 1.17 is the "latest" version officially for whatever spigotmc's reasoning may be)
"1.17.1": triggers the wildcard `elif` statment, installs openjdk16, and builds version 1.17.1
"1.17": triggers it's own statement, installs openjdk 16, and build version 1.17 (not needed now, but may be later for whatever reason someone may want to install such version)
"1.16.5": results in openjdk 8 being installed, and a build of 1.16.5 being done. I would presume any other non-1.17 version would work the same.

I'm sure the script could be cleaner, but it works

### All Submissions:

* [ ] Have you followed the guidelines in our Contributing document?
* [ ] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?
  * If not, why?:

### New Server Submissions:

1. [ ] Have you added an explanation of what your changes do and why you'd like us to include them?


Software-Noob:
Instead of three separate IF/ELIF statements, a better solution would be to future-proof it.  This solution is hard-coded for 1.17 only. Compare the first three numbers ignoring the subversion.

View an [example regex](https://regex101.com/r/G6Kg0p/1) from parker.

PR #1143, opened 2021-05-12
Fix Impostor Server URLs

JohnB17:
### All Submissions:

* [X] Have you followed the guidelines in our Contributing document?
* [X] Have you checked to ensure there aren't other open [Pull Requests](../../pulls) for the same update/change?
* [ ] Did you branch your changes and PR from that branch and not from your master branch?
  * If not, why?: 
It seemed easier.

### Changes to an existing Egg:

1. [X] Have you added an explanation of what your changes do and why you'd like us to include them?
2. [X] Have you tested your Egg changes?

The download URLs changed, this pull request fixes them.

PR #1261, opened 2021-07-22
Update egg-among-us--impostor-server.json

larssieboy18:
Impostor server does not work with the default imagine. Using the image that is listed on their Github (https://github.com/Impostor/Impostor/blob/master/docs/Running-the-server.md), it does work. Might also fix #1143.

### All Submissions:

  * If not, why?:


parkervcp:
I'd rather you update the image versus updating the egg to use your image.

gOOvER:
I see not really any changes in the Image. So what need to be changed?

PR #1263, opened 2021-07-23
Update Technic eggs and Readme

ImZorg:
This PR updates some of the Technic server download URLs to use HTTPS, because HTTP Downloads are no longer supported.

In addition the README was updated to use the new folder structure and include a link to "The 1.12.2 Pack" which was missing.

### All Submissions:

### DIFF
--- a/game_eggs/minecraft/java/technic/README.md
+++ b/game_eggs/minecraft/java/technic/README.md
@@ -1,10 +1,10 @@
 ## Technic Eggs
 
-[Technic](/minecraft/java/technic/)   
-* [Attack of the B-Team](/minecraft/java/technic/attack-of-the-bteam/)
-* [Blightfall](/minecraft/java/technic/blightfall/)
-* [Hexxit](/minecraft/java/technic/hexxit/)  
-* [Tekkit](/minecraft/java/technic/Tekkit/)
-* [Tekkit Classic](/minecraft/java/technic/tekkit-classic/)
-* [Tekkit Legends](/minecraft/java/technic/tekkit-legends/)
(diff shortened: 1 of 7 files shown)


Software-Noob:
I've downloaded using both HTTP and HTTPS without any issues, and their download links use HTTP. Although it doesn't hurt to use HTTPS as that's the standard with redirect support.

It would be preferable to use the panel's editor to modify the JSON files instead of directly touching the file because it's prone to mistakes and doesn't update the date in the file. LGTM, thank you for the contribution 👍

Software-Noob (approved):

ImZorg:
Thank you for your feedback, I will definitely use the panel editor next time. :+1: 

Also, when I use curl to download the server files with the HTTP URL, just like the eggs did, I get the following in the outfile:
```
<html>
<head><title>301 Moved Permanently</title></head>
<body>
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

Maybe it's only on my machine. :shrug: 

Anyways, thank you for responding so quickly.

Software-Noob:
> Thank you for your feedback, I will definitely use the panel editor next time. 👍
> 
> Also, when I use curl to download the server files with the HTTP URL, just like the eggs did, I get the following in the outfile:
> 
> ```
> <html>
> <head><title>301 Moved Permanently</title></head>
> <body>
> <center><h1>301 Moved Permanently</h1></center>
> <hr><center>nginx</center>
> </body>
> </html>
> ```
> 
> Maybe it's only on my machine. 🤷
> 
> Anyways, thank you for responding so quickly.

You should add -L curl flag to follow the redirects which the egg already does

PR #1269, opened 2021-07-25
fix: discord.py generic dependencies not loading on 2.7/3.9

minemalox:
I removed the fixed path on pip install, so that the dependencies are put in their respective folder depending on the python version, instead always being put into ".local/lib/python3.8". 
This PR also fixes the current problem, that dependencies aren't detected by python 3.9

### All Submissions:


parkervcp:
That's the default location python uses. The reason it's used is to guarantee that it's there.

minemalox:
Yeah, found that out while testing the changes out.
The fixed path is somewhat a problem while using the debian_python-3.9 or debian_python-2.7 Images because the dependencies are getting downloaded into ".local/lib/python3.8" instead of ".local/lib/python3.9" or ".local/lib/python2.7" and therefore not loaded by Python 2.7 or 3.9

parkervcp:
I agree but 3.8 is the default image it uses. I really wish there was a better way to do this.

minemalox:
Actually, currently the debian_python-3.9 is the default image for creating a new Server. (At least for me on) 
Or did you mean the installation container?

Maybe it would be better just to set the ".local" directory as a prefix in pip, like I did in my second commit. With that you could be sure that the directory is at the right place and it would work with all python versions. 

Also is there a reason why the packages are installed on install even though they are installed on every startup?

Software-Noob:
> Actually, currently the debian_python-3.9 is the default image for creating a new Server. (At least for me on)

That's something I missed when adding the new images

parkervcp:
Ah I see.

parkervcp (commented):
I can test this out later.

hwalker928:
Just tested it with python 3.9, works flawlessly.

Software-Noob (approved):
This solves the lib path issue, thank you for the contribution. LGTM

PR #1277, opened 2021-07-31
Update fivem_license max input rule to support new key format

LittleDraily:
fix an error with fivem licensekeys that are larger then 32 characters as fivem licensekey length grows

### DIFF
--- a/game_eggs/gta/fivem/egg-five-m.json
+++ b/game_eggs/gta/fivem/egg-five-m.json
@@ -35,7 +35,7 @@
             "default_value": "",
             "user_viewable": true,
             "user_editable": true,
-            "rules": "required|string|max:64"
+            "rules": "required|string|max:33"
         },
         {
             "name": "Max Players",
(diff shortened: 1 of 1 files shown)


parkervcp:
The new keys are a specific length. The should be no reason to go way past that.

Can we get the exact length please.

Software-Noob:
> The new keys are a specific length. The should be no reason to go way past that.
> 
> Can we get the exact length please.

33

game_eggs/gta/fivem/egg-five-m.json
```diff
@@ -35,7 +35,7 @@
             "default_value": "",
             "user_viewable": true,
             "user_editable": true,
-            "rules": "required|string|max:32"
+            "rules": "required|string|max:64"
```
Software-Noob, line 38:
```suggestion
            "rules": "required|string|max:33"
```

Software-Noob (changes requested):

Software-Noob:
I've fixed and tested it on your behalf. 

@parkervcp  I don't know about your current situation with the Fivem team, and as such will leave this up to you

parkervcp:
All this does it fix a var length and doesn't update anything else. Just makes it easier to use the new key length.

Software-Noob (approved):

PR #1319, opened 2021-08-27
Add Satisfactory

TuEye:
### All Submissions:

### New Server Submissions:

2. [x] Have you added the server to the main README.md?
3. [x] Have you added a unique README.md for the server you are adding?


TuEye:
> 
> 
> I honestly would rather wait for an official server than this.

I can understand that of course, neither does the current implementation have the usual tools for administration nor is the experience completely smooth when using it. 

But I am incredibly impatient when it comes to waiting for the "real" dedicated server. I had guessed that others felt the same way. 

I hoped that we could include this Egg until the release of the real server. And would replace it with the proper server after that. After all, this is more functionality than we had before.

If you still prefer to reject the egg (completely understandable), I would have no problem closing the pull request for the now.

parkervcp:
I am going to go with a no here and just stay with that.

PR #1332, opened 2021-09-09
Specify memory for Spigot BuildTools

DAThosting:
Added SERVER_MEMORY to BuildTools command.
Else the install will fail.

1. [ X ] Have you added an explanation of what your changes do and why you'd like us to include them?
2. [ X ] Have you tested your Egg changes?


Software-Noob:
It does not hurt to have this, but building shouldn't and doesn't fail when I test. Are you allocating below 1024 of memory?

DAThosting:
You're right there.
With more than 1024 it works.

Software-Noob (approved):

PR #1337, opened 2021-09-13
Updated Gmod Egg

SenseiHiraku:
### All Submissions:
**Added an option to change the server's version**: Many communities are currently using the 64bit beta version of the game because it offers a newer LuaJIT version and is performing better on some terms, than the stable version.
However, the default egg currently doesn't offer a possibility to change the game's version, that's why I've added one


stock_eggs/source-engine/egg-garrys-mod.json
```diff
@@ -99,6 +99,15 @@
             "user_viewable": true,
             "user_editable": true,
             "rules": "required|boolean"
+        },
+        {
+            "name": "Beta",
+            "description": "Determines which server version should be used. | NONE = stable release | x86-64 = 64Bit beta.\r\nIt is necessary to reinstall the server when changing the version here.",
+            "env_variable": "BETA",
+            "default_value": "NONE",
```
parkervcp, line 107: Is `NONE` a proper value for this to not fail?

parkervcp: I re-read the desc you put in and it looks like it is.

Software-Noob:
We have ${SRCDS_BETAID} for this which the docker image entrypoint would use to auto update on each server startup.

With this, you only install it once and never update correctly again as the entrypoint would install normal branch

stock_eggs/source-engine/egg-garrys-mod.json
SenseiHiraku: Yeah, `NONE` is the value for the stable version of the game.

SenseiHiraku:
> With this, you only install it once and never update correctly again as the entrypoint would install normal branch
Thanks for raising that, I've made adjustments, so it does use the proper env_variable now.

I'm wondering if I should add an extra boolean variable to enable `[VALIDATE](https://github.com/pterodactyl/yolks/blob/4f55fd09d954bba784927e341b9bb96716f90f0f/games/source/entrypoint.sh#L56)` at least once after changing the server's version

brandonsturgeon:
Worth noting that there are some other beta branches, though I'm not sure any of them should be used.

image

+1 on this feature though, this would make transitioning to Pterodactyl much easier.

parkervcp:
> 
> 
> Worth noting that there are some other beta branches, though I'm not sure any of them should be used.
> 
> image
> 
> +1 on this feature though, this would make transitioning to Pterodactyl much easier.

My response to this is that if a host wants to support other features they can modify it.

PR #1343, opened 2021-09-19
Added languagetool server

1Euro7Cent:
### All Submissions:

  * If not, why?:
  I don't know how

### New Server Submissions:


README.md
```diff
@@ -250,3 +250,5 @@ If you are reading this it looks like you are looking to add an egg to your serv
 * [haste-server](/software/haste-server)
 ### RabbitMQ
 * [rabbitmq](/software/rabbitmq)
+### Language-Tool
```
Software-Noob, line 253: The list is in alphabetical order. Put this above RabbitMQ. 

The name is LanguageTool, as seen on their website, not Language-Tool. Please correct this to match in both readmes you've created.

Software-Noob (changes requested):

Software-Noob:
I am not sure if you've tested your new change, but it breaks the egg since it tries to unarchive files without `ngrams` in the filename. The previous version worked fine with the suggestions I provided.

1Euro7Cent:
I tested it. On my site it works. And I dont download with ngrams at the file

Software-Noob:
The download fails, because the host has Cloudflare enabled which blocks the download, and as such only, the Cloudflare HTML file is fetched. You might not have this if you're self-hosting, but most datacenter IPs are flagged causing downloads to fail (403 access denied).

I do see they have a Github for source code, but besides that, this egg will not work if everyone is blocked by the host.

```  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 12409    0 12409    0     0   299k      0 --:--:-- --:--:-- --:--:--  302k
unziping
Archive:  download.zip
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of download.zip or
        download.zip.zip, and cannot find download.zip.ZIP, period.
moving files
cp: can't stat 'LanguageTool-*/*': No such file or directory
```

PR #1382, opened 2021-11-04
[Satisfactory] Removed -multihome flag, no longer required as of patch 0.5.0.7.

Maelstromeous:
### All Submissions:

  * If not, why?:
  * 

### DIFF
--- a/game_eggs/steamcmd_servers/satisfactory/egg-satisfactory.json
+++ b/game_eggs/steamcmd_servers/satisfactory/egg-satisfactory.json
@@ -12,7 +12,7 @@
         "ghcr.io\/pterodactyl\/games:source"
     ],
     "file_denylist": [],
-    "startup": ".\/Engine\/Binaries\/Linux\/UE4Server-Linux-Shipping FactoryGame ?listen -Port={{SERVER_PORT}} -ServerQueryPort={{QUERY_PORT}} -BeaconPort={{BEACON_PORT}} -multihome=0.0.0.0",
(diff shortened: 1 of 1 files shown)


Software-Noob:
It doesn't hurt to keep this ensuring no such breaking occurs again. Similar to how we specify 0.0.0.0 IP in games or applications that support it

Maelstromeous:
@Software-Noob it was always intended as a temporary fix to put  the flag in. Also in its current form IPV6 connectivity does not work.

gOOvER:
> @Software-Noob it was always intended as a temporary fix to put the flag in. Also in its current form IPV6 connectivity does not work.

No, it was not only for Quickfix. This is an official flag for the server and should stay where it is

Maelstromeous:
> > @Software-Noob it was always intended as a temporary fix to put the flag in. Also in its current form IPV6 connectivity does not work.
> 
> No, it was not only for Quickfix. This is an official flag for the server and should stay where it is. You can add also an IPv6 Ip to -multihome

What's the full param? `-multihome=::`?

Software-Noob:
> > > @Software-Noob it was always intended as a temporary fix to put the flag in. Also in its current form IPV6 connectivity does not work.
> > 
> > 
> > No, it was not only for Quickfix. This is an official flag for the server and should stay where it is. You can add also an IPv6 Ip to -multihome
> 
> What's the full param? `-multihome=::`?

That would enable IPV6. From what I've read, leaving this in the startup args doesn't affect anything negatively, and I would rather keep this, for now, similar to how other eggs force 0.0.0.0 to prevent this kind of server breaking. Feel free to open a new PR or issue should the Satisfactory server make new breaking changes.

PR #1407, opened 2021-11-20
[Satisfactory] 0.5.0.12 General QoL Update & EA Readiness

lilkingjr1:
# General Quality of Life Update & Early-Access Readiness

### Egg:
- Added the more up-to-date parkervcp `games:source` image as the primary image, which will hopefully eliminate some edge-case errors people have been having with installing/updating.
- Updated the Configuration Files parser to include entries for new Variables (listed below). Using the "file" parser is only temporary until a patch of the "ini" parser hits the stable branch of Pterodactyl, which will allow proper parsing of Satisfactory's nomenclature with its .ini files.
- Changed the stop command from killing the process to sending a SIGINT, which successfully exits the process gracefully.
- Updated/changed the installation script to accommodate the new Variables listed below. As you can see in the comments, it now currently removes existing .ini files upon re-install to avoid edge-cases of people updating this egg who already have an existing file with certain new variables missing. This would prevent the new variables from working. This method will be revised when the previously mentioned ini parser is updated, but conversely, I do not think it will affect many people negatively in the mean time.
- Added "Autosave Interval" variable.
- Added "Number of Rotating Autosaves" variable.
- Added "Upload Crash Reports" variable.
- Added "[Advanced] Client Initial Connection Timeout" variable.
- Added "[Advanced] Client Established Connection Timeout" variable.
- Added "[Advanced] Network Quality" variable.
- Added "[Advanced] Beta ID" variable.
- Added "[Advanced] Beta Password" variable.

### README:
- Replaced the experimental warning with an Egg update warning. Server is leaving experimental in a week, so it will soon no longer be relevant, and improperly updating the egg will cause more significant issues.
- Adjusted "Egg Capabilities" section.
- Reworked the "Installation/System Requirements" section to be a more detailed & organized with a table.
- Condensed the recommended RAM and Disk sections to the section above.
- Added "Server Initialization" section.
- Updated & improved the "Save Files" section.
- Updated the "List of known commands" link.
- Tried to make the "Errors/Warnings" section more readable.
___
### All Submissions:


Software-Noob:
The games source image won't solve the older kernel issue (kernel below 5) with 32 bit steam networking package as it's more of upstream issue

lilkingjr1:
> The games source image won't solve the older kernel issue (kernel below 5) with 32 bit steam networking package as it's more of upstream issue

Yes, but the `sleep 1` does resolve the issue of a few people I've talked to regardless, so I thought it would still be best to use the image. Also, assuming [my image PR](https://github.com/parkervcp/yolks/pull/20) goes through, it will also fix a new SteamCMD issue found today.

game_eggs/steamcmd_servers/satisfactory/egg-satisfactory.json
```diff
@@ -9,19 +9,20 @@
     "description": "Satisfactory is a first-person open-world factory building game with a dash of exploration and combat. Play alone or with friends, explore an alien planet, create multi-story factories, and enter conveyor belt heaven!",
     "features": null,
     "images": [
+        "ghcr.io\/parkervcp\/games:source",
```
Software-Noob, line 12: Why two different source images?

lilkingjr1:
Hmm... I found a developing issue... On a new server install, Max Players seems to work fine at any value. However, on my existing server, any value still works fine ***unless*** the default value of 4 is used; which then causes the value to be deleted from the file. This means if the value were to ever be changed in the future from the default, it would not work without a re-install. This, among with other difficulties, will not be an issue once the ini parser update is pushed to stable branch, but until then, what do you think is the best approach to mitigate this?

game_eggs/steamcmd_servers/satisfactory/egg-satisfactory.json
lilkingjr1: Freedom of choice ¯\_(ツ)_/¯

Software-Noob: Sometimes a choice isn't necessary, since they both should be doing the same thing;  pushing different changes to one image might cause different outcomes

lilkingjr1: Okay, I'll remove the Pterodactyl image since it tends to be out of date anyways.

Software-Noob:
> Hmm... I found a developing issue... On a new server install, Max Players seems to work fine at any value. However, on my existing server, any value still works fine _**unless**_ the default value of 4 is used; which then causes the value to be deleted from the file by the game server when the file is re-written upon stop. This means if the value were to ever be changed in the future from the default, it would not work without a re-install. This, among with other difficulties, will not be an issue once the ini parser update is pushed to stable branch, but until then, what do you think is the best approach to mitigate this?

Advise users to delete the old config/reinstalling server to update, since there's nothing else we can really do with regards to the difference between the files and different parsers.

Keep in mind that a lot of users don't always update to the latest minor versions, so the INI parser hotfix might not reach everyone right away.

lilkingjr1:
Changes pushed.

> Keep in mind that a lot of users don't always update to the latest minor versions, so the INI parser hotfix might not reach everyone right away.

Has this "hotfix" been published to stable yet? Or do you just mean when it is eventually pushed?

Software-Noob:
> 

The latter. It's going to be in the next Wings release

Software-Noob (approved):
LGTM, thank you for contributing

gOOvER:
for me, the parser don't change values in GameUserSettings.ini.

 "mIntValues": "mIntValues=((\"FG.NetworkQuality\", {{server.build.env.NETWORK_QUALITY}}))"
 
 The File Parser will not find anything with that.

PR #1422, opened 2021-11-27
Fix Rust RCON connection

VibeGAMESNL:
The RCON IP address was not set and has to be 0.0.0.0. Else the rcon connection can't be established using the default pterodactyl networking.

### DIFF
--- a/game_eggs/steamcmd_servers/rust/rust_autowipe/egg-rust-autowipe.json
+++ b/game_eggs/steamcmd_servers/rust/rust_autowipe/egg-rust-autowipe.json
@@ -13,7 +13,7 @@
         "quay.io\/pterodactyl\/core:rust"
     ],
     "file_denylist": [],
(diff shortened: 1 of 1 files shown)


Software-Noob:
This doesn't change anything since it's by default 0.0.0.0 unless configured otherwise. What RCON issue exactly are you trying to solve?

VibeGAMESNL:
> . What RCON issue exactly are you try

Simply if you don't add `+rcon.ip 0.0.0.0` in the startup command, you can't connect to the rcon for some reason. 

After adding +rcon.ip 0.0.0.0, it started to work, removing it broke it again. So doesn't look like Rust sets it to 0.0.0.0 by default, at least not for the rcon.

Software-Noob:
> > . What RCON issue exactly are you try
> 
> Simply if you don't add `+rcon.ip 0.0.0.0` in the startup command, you can't connect to the rcon for some reason. (Tested with [RustAdmin](https://www.rustadmin.com/))
> 
> After adding +rcon.ip 0.0.0.0, it started to work, removing it broke it again. So doesn't look like Rust sets it to 0.0.0.0 by default, at least not for the rcon.

Do you have multiple IPs on your network interface or using host network mode? This shouldn't be a problem nor can I replicate it in a normal setup

VibeGAMESNL:
> ple IPs on your network interface or using host network mode? This shouldn't be a problem nor can

I use the default pterodactyl_nw network. There are indeed 2 IP addresses attached to the main network interface but the Rust server is using the main/default IP.

parkervcp:
I don't see this hurting anything to just add the ip to the server anyways.

parkervcp (approved):

PR #1425, opened 2021-11-29
Add clarification to Red egg readme

AlaRubra123:
When installing the `defender` cog, and possibly others, you'll get the same disk space error as with using the Audio cog, with the same solution.
 
 ### All Submissions:
 * [x]  Have you followed the guidelines in our Contributing document?
 * [x]  Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
 * [x]  Did you branch your changes and PR from that branch and not from your master branch?
 
 1. [x]  Have you added an explanation of what your changes do and why you'd like us to include them?
 2. [x]  Have you tested your Egg changes?

### DIFF
--- a/bots/discord/redbot/README.md
+++ b/bots/discord/redbot/README.md
@@ -8,7 +8,7 @@ No port are required to run Red.
 if you want to use the internal Lavalink Server, you need to allocate port 2333 as primary
 
 ### Additional Requirements
-When using the Audio Cog the bot will attempt to save files to /tmp resulting in a disk space error.  To resolve this error you must increase the size of `tmpfs` using custom container policy.
+When using the Audio Cog the bot will attempt to save files to /tmp resulting in a disk space error.  You may also see this same error when attempting to install a cog, due to pip using /tmp to build the requirements.  To resolve this error you must increase the size of `tmpfs` using custom container policy.
 
 For additional details see: 
 https://pterodactyl.io/wings/1.0/configuration.html#other-values


Software-Noob:
This addition nullifies the previous sentence. There's no reason to mention it twice

AlanBeattyZ:
> This addition nullifies the previous sentence. There's no reason to mention it twice

How does it in any way nullify the previous sentence? The first sentence is regarding using the audio cog. The addition is regarding installing other cogs. It's a different situation in which the same error may arise.

Software-Noob:
At that point it's more clearer to specify cogs, as audio cog would fall into the same cogs category

AlaRubra123:
It's the difference between *using* the audio cog, which is part of
Red's core cogs, and *installing* a different third party cog, using the
downloader cog.

Also, I'm not sure what you mean by "specify cogs", but if you mean
specifying which cogs could lead to the error, that's not really
feasible. Red has over 500 different cogs available on the index, and
probably more besides that. *I* ran into the issue installing defender,
but none of the 12 other non-core cogs I have installed. It could
potentially happen with any of the other cogs on the index that have
additional requirements installed through pip.

Software-Noob:
My message could have been more clear I suppose. For example, proposing a generic notice regarding running out of disk space error, instead of specifying it twice; first explicitly for an audio cog, and then for all other audio cogs.

This is fine too. Thanks :+1:

Software-Noob (approved):

PR #1427, opened 2021-11-29
chore: update Java images for default Minecraft eggs

QuintenQVD0:
### All Submissions:

Most eggs are already update to those images why shouldn't we do the same with this two?
I would like to add this so that if someone downloads the egg from here it will automatically have the correct images.


Software-Noob (approved):

PR #1430, opened 2021-11-29
[Satisfactory] 0.5.1.2 General QoL Update

lilkingjr1:
__**Egg:**__
- Fixed updated variable for uploading crash reports.
- Reworded/Reworked the "Beta ID" startup variable to be more clear and less error-prone to use. I think at this point it is not expected for extra branches to be added or existing branches to be changed, but new branches can easily be added quickly with an egg update in the future.
- Removed the (currently) unused "Beta Password" startup variable.

__**Readme:**__
- Small updates and formatting.

__**Experimental Branch Note:**__
- It appears the `experimental` branch of the server is potentially moving away from a 3 port configuration to a single query port configuration, where all traffic will be passed through. This version of the egg works with both the `public` and `experimental` branches as they currently stand. However, once this change moves to `public`, I will likely submit another egg update to migrate to this new configuration, and the query port will become the new primary port (just as a heads up).

### All Submissions:


Software-Noob (approved):

PR #1500, opened 2021-12-27
Fix "'lib32gcc1' has no installation candidate" error

Craftzman7:
* [x] Have you followed the guidelines in our Contributing document? Yes
* [x] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change? Yes
* [x] Did you branch your changes and PR from that branch and not from your master branch? Yes
  * If not, why?:  N/A

1. [x] Have you added an explanation of what your changes do and why you'd like us to include them? Yes
2. [x] Have you tested your Egg changes? Yes

## Notes

The install script fails to install the "lib32gcc1" package. The replacement package for it is "lib32gcc-s1". The edited script replaces the "lib32gcc1" package" with the "lib32gcc-s1" package.


Software-Noob:
The stock eggs is an attempt to provide a copy of the default panel eggs, although they are outdated most of the time. It's fixed upstream on the panel eggs in a different way and making these different would be confusing

PR #1519, opened 2022-01-10
Update Startup Done Tag to New Behavior

ian-otto:
This change fixes the Risk of Rain 2 server remaining in the STARTING state forever. Due to a recent change in the game server, the game no longer properly executes the substitutions for translations, so the game will only print the untranslated internal string name `CHAT_KEYBIND_TIP`.

I tested these changes within my fresh Pterodactyl installation under a fresh instance to ensure it worked properly.

### All Submissions:

  * If not, why?: N/A


gOOvER:
MB this is only a Bug.

gOOvER:
@Software-Noob This can be closed. There is no need to change this. I test it right now and Server switch to online without problems

gOOvER (changes requested):
Not needed this PR. Work as it should

PR #1529, opened 2022-01-14
Fix for the Download Link

SkyRaax:
### Summary

After some testing I found out that the download link from MohistMC itself changed. The changed isn't this big.
They just removed the file ending "-server.jar" at the end of the link. So the download link should end on ".../download".

(First Pull Request for me here, please say if something is wrong)

### All Submissions:

  * If not, why?:


Software-Noob:
Thank you for your contribution. Please do not edit the JSON files directly and instead use the panel to modify and export eggs. Otherwise, one mistake can break the JSON, and you've left a lot of empty space after the if check.

I am going with #1530 as it exported the egg and contained other changes.

PR #1539, opened 2022-01-23
Mirrors SERVER_MEMORY variable in some MC eggs

SweptWasTaken:
### All Submissions:

  * If not, why?:

I've mirrored the `SERVER_MEMORY` variable to the `-Xms` field in the **Fabric**, **Forge**, **Magma**, **Paper** and **Spigot** MC eggs. However this is because these are the only eggs i've personally tested this with, although I have done a lot of testing with these and am sure it's safe, but if there's more that could benefit from this tweak please let me know and i'll add it to the PR.

I applied this setting after my Minecraft Paper server was crashing with a certain plugin enabled. After I applied this tweak I haven't gotten a crash since. I looked around for reasons why this already wasn't the case in Pterodactyl but couldn't find any so forgive me if this has already been discussed.

Credit to [Aikar's Java Flags](https://blog.airplane.gg/aikar-flags/).

### DIFF
--- a/game_eggs/minecraft/java/fabric/egg-fabric.json
+++ b/game_eggs/minecraft/java/fabric/egg-fabric.json
@@ -19,7 +19,7 @@
         "ghcr.io\/pterodactyl\/yolks:java_8"
     ],
     "file_denylist": [],
-    "startup": "java -Xms128M -Xmx{{SERVER_MEMORY}}M -jar {{SERVER_JARFILE}}",
+    "startup": "java -Xms{{SERVER_MEMORY}}M -Xmx{{SERVER_MEMORY}}M -jar {{SERVER_JARFILE}}",
     "config": {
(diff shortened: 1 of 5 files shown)


yesBad:
It would seem you didnt read Aikar's flags clearly enough.

SweptWasTaken:
> It would seem you didnt read Aikar's flags clearly enough.

Could you point out something i've missed? This is the original article which explains that having your `-Xms` and `-Xmx` being the same is beneficial.

https://aikar.co/mcflags.html

Software-Noob:
There's no reason to force the minimum heap to use all the available memory right away for everyone by default. 

Aikar flags do suggest this, but they modify a lot of other aspects besides xms. They also have special instructions regarding Pterodactyl, because of allocating everything to Java will starve the container and OOM it.

PR #1541, opened 2022-01-24
[curseforge-generic] Fix curl not accepting URLs with brackets

AestasLonewolf:
Fix for curseforge-generic when modpack name contains brackets


Software-Noob (approved):

Issue #1573, opened 2022-03-02
[Bug]: nanolimbo egg seems to be broken


torunoya:
### Panel Version

1.7.0

### Wings Version

1.6.1

### Service

minecraft/java/nanolimbo

### Modified

No, I did not modify the egg

### Expected Behavior

install the server, start it, play

### Actual Behavior

install the server, start it, 

[ERROR]: Cannot start server: 
java.lang.NullPointerException
        at io.netty.buffer.ByteBufUtil.utf8Bytes(ByteBufUtil.java:1080)
        at ru.nanit.limbo.protocol.ByteMessage.writeString(ByteMessage.java:95)
        at ru.nanit.limbo.protocol.packets.play.PacketPlayerInfo.encode(PacketPlayerInfo.java:52)
        at ru.nanit.limbo.protocol.PacketSnapshot.encode(PacketSnapshot.java:51)
        at ru.nanit.limbo.protocol.PacketSnapshot.of(PacketSnapshot.java:86)
        at ru.nanit.limbo.connection.ClientConnection.initPackets(ClientConnection.java:442)
        at ru.nanit.limbo.server.LimboServer.start(LimboServer.java:73)
        at ru.nanit.limbo.NanoLimbo.main(NanoLimbo.java:27)

### Steps To Reproduce

Step 1 Set version to latest
Step 2 install
Step 3 start it
Step 4 Receive error "java.lang.NullPointerException
        at io.netty.buffer.ByteBufUtil.utf8Bytes(ByteBufUtil.java:1080)
        at ru.nanit.limbo.protocol.ByteMessage.writeString(ByteMessage.java:95)
        at ru.nanit.limbo.protocol.packets.play.PacketPlayerInfo.encode(PacketPlayerInfo.java:52)
        at ru.nanit.limbo.protocol.PacketSnapshot.encode(PacketSnapshot.java:51)
        at ru.nanit.limbo.protocol.PacketSnapshot.of(PacketSnapshot.java:86)
        at ru.nanit.limbo.connection.ClientConnection.initPackets(ClientConnection.java:442)
        at ru.nanit.limbo.server.LimboServer.start(LimboServer.java:73)
        at ru.nanit.limbo.NanoLimbo.main(NanoLimbo.java:27)"

### Install logs

command is taking hours to load

torunoya:
[install.log](https://github.com/parkervcp/eggs/files/8168791/install.log)

NgLoader:
[...]

PR #1574, opened 2022-03-02
fix: fixed nanolimbo now pulling latest config

NgLoader:

The install script will now pull the latest settings.yml from github

fix: #1573


Software-Noob:
Do not pull configs from remote sources as there can be any breaking change there. It should be written locally as it is right now and updated if something is missing, or uploaded to the repository

Software-Noob (approved):

PR #1608, opened 2022-04-11
Fix spongeforge install

brandonsturgeon:
### There are three issues with the existing script:
---
 **1. Some of the `curl` commands use the `-o` flag twice**
For example: `curl -s -o server.jar -o https://blah.blah`.
`-o` should only be used to direct the output to the final `.jar` file. The second `-o` is erroneous.
```diff
-- curl -s -o server.jar -o {{THE_URL}}
++ curl -s -o server.jar {{THE_URL}}
```
---
 **2. The CDN shipping these jar files has changed things up a bit.**
The download links now redirect to a different URL. To accommodate this change, I added the `-L` flag on the `curl` commands (follow redirects)'
```diff
-- curl -s -o server.jar {{THE_URL}}
++ curl -Ls -o server.jar {{THE_URL}}
```
---
  **3. There's an `mv` command that moves things weirdly.**
It moves the existing `spongeforge.jar` file(s?) into `mods/old/`, but the `mv` command forcefully renames the file to `mods/old/spongeforge*.jar`.
I changed it to be like: `mv -vf spongeforge*.jar mods/old/`.
This allows the command to work with multiple `spongeforge*.jar` files (if that'd ever happen), but more importantly it'll also retain its filename after being moved. I also added the `-v` flag just for some better debug info.
```diff
-- mv -f /mnt/server/mods/spongeforge*.jar /mnt/server/mods/old/spongeforge*.jar
++ mv -vf /mnt/server/mods/spongeforge*.jar /mnt/server/mods/old/
```
---

These changes fix all of those problems, allowing the installation script to complete.

### ⚠️ However ⚠️
There is still an issue with the install script that I wasn't able to figure out. Somewhere along the way, the install script will create the final server jar, but it won't be named `server.jar`. This seems to be the result of one of the installer jars, not our install script.

For example, here's my filesystem after running this new installer:
image

`server.jar` will fail on startup, but `minecraft_server.1.12.2.jar` succeeds.

### 💡 My Suggestion 💡
I think it would be safe to do a simple `rm -v server.jar && mv minecraft_server.*.jar server.jar` after the script succeeds. That would make sure the `server.jar` is correct.

The reason I didn't include that in this PR is because there may be a more elegant or "correct" way to address this, but I think that's out of my purview.

### All Submissions:


parkervcp:
@brandonsturgeon Why not download the installer to an `install.jar` and then just move the server jar to the final `server.jar`

PR #1610, opened 2022-04-16
Update egg-discord-java.json

James-Discord:
Imagine advertising in a egg... can't be sneaky hub the scummiest company/one of the scummiest on hosting

### DIFF
--- a/bots/discord/discord.java/egg-discord-java.json
+++ b/bots/discord/discord.java/egg-discord-java.json
@@ -6,7 +6,7 @@
     },
     "exported_at": "2021-07-21T18:23:17+03:00",
     "name": "Discord Java",
-    "author": "sneaky@sneakyhub.com",
+    "author": "parkervcp@pterodactyl.io",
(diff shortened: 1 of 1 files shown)


gOOvER:
Why you change the Author?

James-Discord:
because why not, its basically a walking advertisement for a company that runs crypto miners and has its own fuckin money that is worth more than 1 dollar

gOOvER:
> because why not, its basically a walking advertisement for a company that runs crypto miners and has its own fuckin money that is worth more than 1 dollar

he wrote the egg, so he's the Author. No need to change it. If you dont like him, then thats your problem

James-Discord:
okay whatever u say

gOOvER (changes requested):
You can't change the Author

Software-Noob:
This has nothing to do with advertising as it's the app author that's set and used for all new eggs. It's a way to show who created the egg.

PR #1630, opened 2022-05-03
added csgo egg

SanjaySRocks:
I have tested the egg and it works fine...


Boy132:
There is already a [cs:go egg](https://github.com/parkervcp/eggs/blob/master/stock_eggs/source-engine/egg-counter--strike--global-offensive.json). (it's a stock/ default egg)

Software-Noob:
Csgo is a default egg shipped with the Pterodactyl panel, and can also be found under the stock folder of this repository

SanjaySRocks:
my bad thanks. I would like to update that stock egg with new update.

PR #1643, opened 2022-05-08
fix: PostgresSQL

### LINKED ISSUE #1549: [Bug]: Postgres egg seems to be broken
### Panel Version

1.7.0

### Wings Version

1.6.1

### Service

postgres

### Modified

No, I did not modify the egg

### Expected Behavior

After setting up the server i wanted to connect to the postgres db.

### Actual Behavior

Trying to connect via pgadmin results in "FATAL: password authentication failed for user 'pterodactyl'"

### Steps To Reproduce

1. Download the egg and create a port allocation in the node
2. Create a server using the postgres egg and change the default passwords
3. Replace postgresql.conf with the provided 
4. Add the pg_hba.conf entry for the host to be able to connect from
5. Try to connect to the database with the configured password

### Install logs

command is taking hours to load smh

TuEye:
* Replaced image with the yolk one
* Made the superuser name configurable
* Removed unused variables
* Removed unneeded installer lines
* Remove unused postgres.conf
   * I have really not been able to figure out why this file is still needed
* Closes #1549

### All Submissions:


Software-Noob:
Thank you for your contribution. You'll have to configure `pg_hba.conf` or no remote connections will be allowed, even if you modify the listen_address. By default, only localhost is whitelisted `127.0.0.1/32`. Change it to `0.0.0.0/0`

TuEye:
> Thank you for your contribution. You'll have to configure `pg_hba.conf` or no remote connections will be allowed, even if you modify the listen_address. By default, only localhost is whitelisted `127.0.0.1/32`. Change it to `0.0.0.0/0`.

I wasn't sure if I should include this change in the Egg. After all, it makes the database immediately accessible from everywhere. 

The advantage of this is, of course, that the Egg immediately offers a turnkey database. I have now included it.

Software-Noob:
> > Thank you for your contribution. You'll have to configure `pg_hba.conf` or no remote connections will be allowed, even if you modify the listen_address. By default, only localhost is whitelisted `127.0.0.1/32`. Change it to `0.0.0.0/0`.
> 
> I wasn't sure if I should include this change in the Egg. After all, it makes the database immediately accessible from everywhere. 
> 
> The advantage of this is, of course, that the Egg immediately offers a turnkey database. I have now included it.
> 
> 

There's a forced authentication with the use of md5 password. Users can specify availability of the server using allocations as with all servers. Generally all servers of Pterodactyl are available right away.

The only other possible thing is to force a minimum password length in the input variable rules. At some point in Pterodactyl V2 there will be a way to randomize the values.

DTS-11:
umm will this be merged soon?

TuEye:
> umm will this be merged soon?

Take the updated egg from this pr and test if it works for you. If it does, just write it here as a comment. This increases the likelihood of a merge dramatically ;-)

database/sql/postgres/egg-postgres.json
```diff
@@ -22,7 +22,7 @@
     },
     "scripts": {
         "installation": {
-            "script": "#! \/bin\/ash\r\nadduser -D -h \/home\/container container\r\n\r\nchown -R container: \/mnt\/server\/\r\n\r\nsu container -c 'initdb -D \/mnt\/server\/postgres_db\/ -A md5 -U \"$PGUSER\" --pwfile=<(echo \"$PGPASSWORD\")'\r\n\r\nmkdir -p \/mnt\/server\/postgres_db\/run\/\r\n\r\necho -e \"Done\"",
+            "script": "#! \/bin\/ash\r\nadduser -D -h \/home\/container container\r\n\r\nchown -R container: \/mnt\/server\/\r\n\r\nsu container -c 'initdb -D \/mnt\/server\/postgres_db\/ -A md5 -U \"$PGUSER\" --pwfile=<(echo \"$PGPASSWORD\")'\r\n\r\nmkdir -p \/mnt\/server\/postgres_db\/run\/\r\n\r\n## Add default \"allow from all\" auth rule to pg_hba\r\nif ! grep -q \"# Custom rules\" \"\/mnt\/server\/postgres_db\/pg_hba.conf\"; then\r\n    echo -e \"# Custom rules\\nhost    all             all             0.0.0.0\/0               md5\" >> \"\/mnt\/server\/postgres_db\/pg_hba.conf\"\r\nfi\r\n\r\necho -e \"Done\"",
```
Software-Noob, line 25: This should be in the config parser. It runs on each server startup instead of only on installation. Other than that, LGTM.

Software-Noob (changes requested):

database/sql/postgres/egg-postgres.json
TuEye: This is a little more difficult than I originally thought.
The only posibility I would have would be to search for "host all all" with the file parser and then replace it with the wanted content (e.g. "host all all 0.0.0.0/0 md5"). 

But even if I put the address into a variable and make it configurable, I would prevent the user from configuring multiple entries to different IP addresses, e.g:
 > host all all 5.5.5/32 md5
 > host all all 4.4.4.4/32 md5

Of course, I don't know if such a customisation would even be widely used ^^
If this restriction is not a problem, I can adjust the Egg accordingly.

DTS-11:
where are the db connection credentials located, like connection url etc?

Software-Noob (approved):

MisterGranti67:
The problem is solved, I used the default accesses given in the egg, and everything works.
Don't forget that the default database is "postgres".

PR #1654, opened 2022-05-15
Add Egg to Support Passing Token Through Command Line

FlopsiBunny:

1. [Y ] Have you added an explanation of what your changes do and why you'd like us to include them?
2. [Y ] Have you tested your Egg changes?


Software-Noob:
A whole new egg for a single variable seems excessive. If you intended to modify the existing egg instead of creating a completely new one, in that case; it's a generic python script egg, and forcing tokens is unnecessary.

Although it's under "discord", it's not restricted to that alone and we had plans to rename the eggs a while back.

Issue #1662, opened 2022-05-18
[Egg Request]: V Rising


MADDMAXKZ:
### Does this expand an already existing service

No

### Link to a game or software

V Rising Egg

### Links for the download

N/A

### Links for the install docs

Requesting a Game egg For Pterodactyl

gOOvER:
https://github.com/StunlockStudios/vrising-dedicated-server-instructions/issues/1#issuecomment-1129496258

mutschler:
Here's a Startingpoint:

It works but Im not really sure if all the commands are supported via commandline, so you'll most likely need to edit the `ServerHostSettings.json` manually after install..

```
{
    "_comment": "DO NOT EDIT: FILE GENERATED AUTOMATICALLY BY PTERODACTYL PANEL - PTERODACTYL.IO",
    "meta": {
        "version": "PTDL_v1",
        "update_url": null
    },
    "exported_at": "2022-05-18T10:04:22+02:00",
    "name": "V Rising",
    "author": "unknown@unknown.com",
    "description": "V Rising Server see https:\/\/github.com\/StunlockStudios\/vrising-dedicated-server-instructions",
    "features": null,
    "images": [
        "quay.io\/parkervcp\/pterodactyl-images:debian_wine-5"
    ],
    "file_denylist": [],
    "startup": "xvfb-run wine .\/VRisingServer.exe -nographics -batchmode -serverName \"{{SERVER_NAME}}\" -port {{SERVER_PORT}} -password \"{{SERVER_PASSWORD}}\" -listOnMasterServer {{SERVER_PUBLIC}} -maxConnectedUsers {{MAX_USERS}}",
    "config": {
        "files": "{}",
        "startup": "{\r\n    \"done\": \"Loaded ServerGameSettings:\",\r\n\"userInteraction\": []\r\n}",
        "logs": "{}",
        "stop": "^V"
    },
    "scripts": {
        "installation": {
[...]

PR #1680, opened 2022-05-28
Add V Rising Egg

lilkingjr1:
Egg made in conjunction with @kapatheus. Resolves egg request issue #1662.

### All Submissions:

### New Server Submissions:

### DIFF
--- a/game_eggs/steamcmd_servers/v_rising/egg-v-rising.json
+++ b/game_eggs/steamcmd_servers/v_rising/egg-v-rising.json
@@ -1,0 +1,231 @@
+{
+    "_comment": "Pterodactyl V Rising Egg ~ David Wolfe (Red-Thirten) and Kapatheus ~ 2022-05-26",
+    "meta": {
+        "version": "PTDL_v1",
+        "update_url": null
+    },
+    "name": "V Rising",
+    "author": "rehlmgaming@gmail.com",
+    "description": "Awaken as a vampire. Hunt for blood in nearby settlements to regain your strength and evade the scorching sun to survive. Raise your castle and thrive in an ever-changing open world full of mystery. Gain allies online and conquer the land of the living.",
+    "features": [
+        "steam_disk_space"
+    ],
+    "images": [
+        "ghcr.io\/parkervcp\/yolks:wine_latest"
+    ],
+    "file_denylist": [],
+    "startup": "wine64 .\/VRisingServer.exe -persistentDataPath save-data -address 0.0.0.0",
+    "config": {
+        "files": "{\r\n\t\"save-data\/Settings\/ServerHostSettings.json\": {\r\n\t\t\"parser\": \"json\",\r\n\t\t\"find\": {\r\n\t\t\t\"Name\": \"{{server.build.env.SERVER_NAME}}\",\r\n\t\t\t\"Description\": \"{{server.build.env.DESCRIPTION}}\",\r\n\t\t\t\"Port\": \"{{server.build.default.port}}\",\r\n\t\t\t\"QueryPort\": \"{{server.build.env.QUERY_PORT}}\",\r\n\t\t\t\"MaxConnectedUsers\": \"{{server.build.env.MAX_PLAYERS}}\",\r\n\t\t\t\"MaxConnectedAdmins\": \"{{server.build.env.MAX_ADMINS}}\",\r\n\t\t\t\"ServerFps\": \"{{server.build.env.FPS}}\",\r\n\t\t\t\"SaveName\": \"{{server.build.env.SAVE_NAME}}\",\r\n\t\t\t\"Password\": \"{{server.build.env.SERVER_PASSWORD}}\",\r\n\t\t\t\"Secure\": \"{{server.build.env.SERVER_SECURE}}\",\r\n\t\t\t\"ListOnMasterServer\": \"{{server.build.env.SERVER_LIST}}\",\r\n\t\t\t\"AutoSaveCount\": \"{{server.build.env.SAVE_COUNT}}\",\r\n\t\t\t\"AutoSaveInterval\": \"{{server.build.env.SAVE_INTERVAL}}\",\r\n\t\t\t\"GameSettingsPreset\": \"{{server.build.env.GAME_SETTINGS_PRESET}}\",\r\n\t\t\t\"AdminOnlyDebugEvents\": \"{{server.build.env.ADMIN_ONLY_DEBUG_EVENTS}}\",\r\n\t\t\t\"DisableDebugEvents\": \"{{server.build.env.DEBUG_EVENTS}}\",\r\n\t\t\t\"Rcon.Enabled\": \"{{server.build.env.RCON}}\",\r\n\t\t\t\"Rcon.Password\": \"{{server.build.env.RCON_PASS}}\",\r\n\t\t\t\"Rcon.Port\": \"{{server.build.env.RCON_PORT}}\"\r\n\t\t}\r\n\t}\r\n}",
+        "startup": "{\r\n    \"done\": \"Loaded ServerGameSettings\"\r\n}",
+        "logs": "{}",
+        "stop": "^C"
+    },
+    "scripts": {
+        "installation": {
+            "script": "#!\/bin\/bash\r\n\r\n## File: Pterodactyl V Rising Egg - egg-v-rising.json\r\n## Authors: David Wolfe (Red-Thirten), Kapatheus\r\n## Date: 2022\/05\/26\r\n## License: MIT License\r\n## Image to install with is 'ghcr.io\/parkervcp\/installers:debian'\r\n\r\n# Install packages. Default packages below are skipped if using image noted above, thus speeding up the install process.\r\napt -y update\r\napt -y --no-install-recommends install curl lib32gcc-s1 ca-certificates\r\n\r\n# Download and install SteamCMD\r\nexport HOME=\/mnt\/server\r\ncd \/tmp\r\nmkdir -p $HOME\/steamcmd $HOME\/steamapps\r\ncurl -sSL -o steamcmd.tar.gz https:\/\/steamcdn-a.akamaihd.net\/client\/installer\/steamcmd_linux.tar.gz\r\ntar -xzvf steamcmd.tar.gz -C $HOME\/steamcmd\r\ncd $HOME\/steamcmd\r\n\r\n# SteamCMD fails otherwise for some reason, even running as root.\r\n# This is changed at the end of the install process anyways.\r\nchown -R root:root \/mnt\r\n\r\n# Install game server using SteamCMD\r\n.\/steamcmd.sh +force_install_dir $HOME +login anonymous $( [[ \"${WINDOWS_INSTALL}\" == \"1\" ]] && printf %s '+@sSteamCmdForcePlatformType windows' ) +app_update ${SRCDS_APPID} validate +quit\r\n\r\n# Set up 32 and 64 bit libraries\r\nmkdir -p $HOME\/.steam\/sdk{32,64}\r\ncp -v linux32\/steamclient.so $HOME\/.steam\/sdk32\/steamclient.so\r\ncp -v linux64\/steamclient.so $HOME\/.steam\/sdk64\/steamclient.so\r\n\r\n## V Rising Setup\r\nmkdir -p $HOME\/save-data\/Settings && cd \"$_\"\r\ntouch adminlist.txt banlist.txt\r\ncat > ServerHostSettings.json << EOF\r\n{\r\n  \"Name\": \"${SERVER_NAME}\",\r\n  \"Description\": \"${DESCRIPTION}\",\r\n  \"Port\": ${SERVER_PORT},\r\n  \"QueryPort\": ${QUERY_PORT},\r\n  \"MaxConnectedUsers\": ${MAX_PLAYERS},\r\n  \"MaxConnectedAdmins\": ${MAX_ADMINS},\r\n  \"ServerFps\": ${FPS},\r\n  \"SaveName\": \"${SAVE_NAME}\",\r\n  \"Password\": \"${SERVER_PASSWORD}\",\r\n  \"Secure\": ${SERVER_SECURE},\r\n  \"ListOnMasterServer\": ${SERVER_LIST},\r\n  \"AutoSaveCount\": ${SAVE_COUNT},\r\n  \"AutoSaveInterval\": ${SAVE_INTERVAL},\r\n  \"GameSettingsPreset\": \"${GAME_SETTINGS_PRESET}\",\r\n  \"AdminOnlyDebugEvents\": ${ADMIN_ONLY_DEBUG_EVENTS},\r\n  \"DisableDebugEvents\": ${DEBUG_EVENTS},\r\n  \"Rcon\": {\r\n    \"Enabled\": ${RCON},\r\n    \"Password\": \"${RCON_PASS}\",\r\n    \"Port\": ${RCON_PORT}\r\n  }\r\n}\r\nEOF",
+            "container": "ghcr.io\/pterodactyl\/installers:debian",
+            "entrypoint": "\/bin\/bash"
+        }
+    },
+    "variables": [
+        {
+            "name": "[REQUIRED] Server Query Port",
+            "description": "UDP port for Steam server list features.",
+            "env_variable": "QUERY_PORT",
+            "default_value": "9877",
+            "user_viewable": false,
+            "user_editable": false,
+            "rules": "required|integer|between:1024,65536"
+        },
+        {
+            "name": "Automatic Updates",
+            "description": "Quickly checks for any server updates on startup, and updates if necessary.",
+            "env_variable": "AUTO_UPDATE",
+            "default_value": "1",
+            "user_viewable": true,
+            "user_editable": true,
+            "rules": "boolean"
+        },
+        {
+            "name": "Game Mode",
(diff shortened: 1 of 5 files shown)


game_eggs/steamcmd_servers/v_rising/egg-v-rising.json
```diff
@@ -0,0 +1,231 @@
...
+            "rules": "required|integer|min:1"
+        },
+        {
+            "name": "List On Master Server",
+            "description": "Set to true to list on the server list, else set to false.",
+            "env_variable": "SERVER_LIST",
+            "default_value": "true",
+            "user_viewable": true,
+            "user_editable": true,
+            "rules": "required|string|in:true,false"
```
gOOvER, line 138: This will save true in a wrong way.

it should be:

`"ListOnMasterServer": true,`

and with this it is

"ListOnMasterServer": "true",


Then the config don't work

game_eggs/steamcmd_servers/v_rising/egg-v-rising.json
```diff
@@ -0,0 +1,231 @@
...
+            "rules": "boolean"
+        },
+        {
+            "name": "Game Mode",
+            "description": "What preset of game settings the server should run. Some settings may not apply after the save file is first created. Leave empty\/null if you are using a single-player uploaded save or a custom ServerGameSettings.json file.",
+            "env_variable": "GAME_SETTINGS_PRESET",
+            "default_value": "StandardPvP",
+            "user_viewable": true,
+            "user_editable": true,
+            "rules": "nullable|string|in:,DuoPvP,HardcorePvP,Level30PvE,Level30PvP,Level50PvE,Level50PvP,Level70PvE,Level70PvP,SoloPvP,StandardPvE_Easy,StandardPvE_Hard,StandardPvE,StandardPvP_Easy,StandardPvP_Hard,StandardPVP"
```
gOOvER, line 57: Gamemode is only "PvP" or "PvE"

gOOvER, line 57: Game mode is only PvP or PvE
"GameModeType": "PvE",

Looks like you mixed up the variables

game_eggs/steamcmd_servers/v_rising/egg-v-rising.json
```diff
@@ -0,0 +1,231 @@
...
+    "author": "rehlmgaming@gmail.com",
+    "description": "Awaken as a vampire. Hunt for blood in nearby settlements to regain your strength and evade the scorching sun to survive. Raise your castle and thrive in an ever-changing open world full of mystery. Gain allies online and conquer the land of the living.",
+    "features": [
+        "steam_disk_space"
+    ],
+    "images": [
+        "ghcr.io\/parkervcp\/yolks:wine_latest"
+    ],
+    "file_denylist": [],
+    "startup": "wine64 .\/VRisingServer.exe -persistentDataPath save-data -address 0.0.0.0",
```
gOOvER, line 17: you dont need wine64, but you should start it with xvfb-run -a. Also you can overwrite a lot of variables

`xvfb-run -a wine ./VRisingServer.exe -serverName "{{SERVER_NAME}}"  -saveName "{{SAVE_NAME}}" -persistentDataPath ./server-data/ -maxConnectedUsers {{MAX_USERS}} -maxConnectedAdmins {{MAX_ADMINS}}`

game_eggs/steamcmd_servers/v_rising/egg-v-rising.json
lilkingjr1: `"true"` and `"false"` happen to work for the RCON setting in the same file (I tested it), so I assume the JSON parser in this particular server application is okay with string booleans (unlike Arma Reforger, for example). Do you know for certain that it will not register properly for the `ListOnMasterServer` setting specifically? It's hard to test this setting reliably given the reports that servers sometimes don't show on the server in general.

gOOvER: thats the problem in my VRising egg and thats why i open  https://github.com/pterodactyl/panel/issues/4087
with "true", true is not recocnized.

"Setting" true, -> ok
"Setting" "true", not possible

game_eggs/steamcmd_servers/v_rising/egg-v-rising.json
lilkingjr1: This is not true. Please read the [documentation](https://github.com/StunlockStudios/vrising-dedicated-server-instructions#configuring-the-server). Current valid values are listed after the `in:` rule.

game_eggs/steamcmd_servers/v_rising/egg-v-rising.json
lilkingjr1: > you dont need wine64. you can use wine

Everything I read online implies 64-bit applications should be run with with `wine64`. Is there a benefit to using `wine`?

> xvfb-run -a

What does this do exactly, and why is it needed? The server seemingly runs fine without it.

> Also you can overwrite a lot of variables

Yes, but I found it ultimately misleading for the end user if they ever open up `ServerHostSettings.json` to take a look. The startup flags do not cause their values to be written to the `ServerHostSettings.json` file, and there are some settings in the file that cannot be set by flags. So, ultimately, I felt it was cleaner to have every setting consolidated to the file instead of disjointed.

gOOvER: you also should add a dir in save-data called Settings

p.e.
`mkdir -p $HOME/server-data/Settings`

with `persistentDataPath` the server searches for the config in `$HOME/server-data/Settings`

take care about: https://github.com/StunlockStudios/vrising-dedicated-server-instructions/issues/26

game_eggs/steamcmd_servers/v_rising/egg-v-rising.json
lilkingjr1: Why does it work for the RCON setting then? It wouldn't make sense for different settings in the file to be parsed differently. Have you considered the possibility that your server isn't in the serverlist due to a factor other than the config file (like server popularity, or master server lag)? Also, the default value for that setting is true according to documentation, so even if it thinks `"true"` is invalid, it would default to `true` anyways.

game_eggs/steamcmd_servers/v_rising/egg-v-rising.json
gOOvER: You misunderstand me;

GameMode and GameModePresents are two different configuration settings.

So the name of the variable is wrong

gOOvER: [Game-Server-Settings.pdf](https://github.com/parkervcp/eggs/files/8790745/Game-Server-Settings.pdf)

This is better for GameSettings :)

game_eggs/steamcmd_servers/v_rising/egg-v-rising.json
lilkingjr1: > you also should add a dir in save-data called Settings
> take care about: https://github.com/StunlockStudios/vrising-dedicated-server-instructions/issues/26

Please stop leaving incorrect comments that clearly indicate you have not read the egg in the PR. The egg's installation script already creates this directory correctly. The issue linked was user error of them putting the `adminlist.txt` file within the `server-data` directory when it should have been placed in the `server-data/Settings` directory (which this egg does correctly.

gOOvER: admin.txt does not  work in  persistentDataPath. Thats what i want to say ;)

game_eggs/steamcmd_servers/v_rising/egg-v-rising.json
lilkingjr1: The startup parameter named "Game Mode" sets the `GameSettingsPreset` setting in ServerHostSettings.json, **not** `GameMode` in ServerGameSettings.json. It was named "Game Mode" to be more recognizable than "Game Settings Preset" to the layman, but if just the name itself causes confusion I can change it. Advanced users are welcome to null this setting and use a custom ServerGameSettings.json file if they would like using the document you linked, but you already know this because the README states all of this.

gOOvER:
```
## V Rising Setup
mkdir -p $HOME/save-data/Settings && cd "$_"
touch adminlist.txt banlist.txt
cat > ServerHostSettings.json << EOF
{
  "Name": "${SERVER_NAME}",
  "Description": "${DESCRIPTION}",
  "Port": ${SERVER_PORT},
  "QueryPort": ${QUERY_PORT},
  "MaxConnectedUsers": ${MAX_PLAYERS},
  "MaxConnectedAdmins": ${MAX_ADMINS},
  "ServerFps": ${FPS},
  "SaveName": "${SAVE_NAME}",
  "Password": "${SERVER_PASSWORD}",
  "Secure": ${SERVER_SECURE},
  "ListOnMasterServer": ${SERVER_LIST},
  "AutoSaveCount": ${SAVE_COUNT},
  "AutoSaveInterval": ${SAVE_INTERVAL},
  "GameSettingsPreset": "${GAME_SETTINGS_PRESET}",
  "AdminOnlyDebugEvents": ${ADMIN_ONLY_DEBUG_EVENTS},
  "DisableDebugEvents": ${DEBUG_EVENTS},
  "Rcon": {
    "Enabled": ${RCON},
    "Password": "${RCON_PASS}",
    "Port": ${RCON_PORT}
  }
}
EOF
```
`ServerGameSettings.json` should also be in `persistentDataPath`

game_eggs/steamcmd_servers/v_rising/egg-v-rising.json
gOOvER: Default is:
`"GameSettingsPreset": "",`

lilkingjr1:
> ServerGameSettings.json should also be in persistentDataPath

Only if you want to override the `GameSettingsPreset` setting or a `ServerGameSettings.json` file stored in the save file (in case of an uploaded save file). I feel having the "override" installed by default is not the optimal default, and it should be left to advanced users to create if they do not want to use an easy predefined Game Settings Preset.

game_eggs/steamcmd_servers/v_rising/egg-v-rising.json
lilkingjr1: True, but assuming a `ServerGameSettings.json` override file is not present, a `null` value will default to the "default" `ServerGameSettings.json` file, which is a copy of `StandardPvP.json`, so it's effectively the same. We can change the default value for this setting to `null` if you would like, but it really doesn't make a difference.

game_eggs/steamcmd_servers/v_rising/egg-v-rising.json
lilkingjr1: Worst case scenario I can change it to file parser, but it seems to be working for me for this specific server application.

game_eggs/steamcmd_servers/v_rising/egg-v-rising.json
gOOvER: i'll do some tests atm:

for now:
RCON "true"; -> works
Secure "true" -> works

gOOvER: I used the same rule before the last patch and it definitely didn't work. There is also a conversation on their Discord somewhere, because I had asked why it does not work with "VALUE", but only with VALUE.

It seems as if they have fixed this with the last patch quietly :)

"ListOnMasterServer": "true", -> works

game_eggs/steamcmd_servers/v_rising/egg-v-rising.json
gOOvER: > Worst case scenario I can change it to file parser, but it seems to be working for me for this specific server application.

I used the same rule before the last patch and it definitely didn't work. There is also a conversation on their Discord somewhere, because I had asked why it does not work with "VALUE", but only with VALUE.

It seems as if they have fixed this with the last patch quietly :)

"ListOnMasterServer": "true", -> works

RCON "true", -> works
Secure "true",-> works

game_eggs/steamcmd_servers/v_rising/egg-v-rising.json
```diff
@@ -0,0 +1,231 @@
...
+            "default_value": "1",
+            "user_viewable": true,
+            "user_editable": true,
+            "rules": "boolean"
+        },
+        {
+            "name": "Game Mode",
+            "description": "What preset of game settings the server should run. Some settings may not apply after the save file is first created. Leave empty\/null if you are using a single-player uploaded save or a custom ServerGameSettings.json file.",
+            "env_variable": "GAME_SETTINGS_PRESET",
+            "default_value": "StandardPvP",
```
PsychoZander, line 54: default_value is StandardPvP and the rules specify StandardPVP, meaning the default does not work out of the box. Not sure which is correct, but the server has started with StandardPVP as the setting.

gOOvER: > default_value is StandardPvP and the rules specify StandardPVP, meaning the default does not work out of the box. Not sure which is correct, but the server has started with StandardPVP as the setting.

He start, but after a clean install it's "". As i wrote above. Not everyone want to use presents. Presents should be optional

game_eggs/steamcmd_servers/v_rising/egg-v-rising.json
lilkingjr1: Commit [bd77d26](https://github.com/parkervcp/eggs/pull/1680/commits/bd77d26939a8b18887771d0d52fb0d4164996c93) should resolve this issue.

game_eggs/steamcmd_servers/v_rising/egg-v-rising.json
lilkingjr1: My apologies. You are correct in that `adminlist.txt` and `banlist.txt` are not currently recognized from within the `~/save-data/` or `~/save-data/Settings/` directories. Commit [bd77d26](https://github.com/parkervcp/eggs/pull/1680/commits/bd77d26939a8b18887771d0d52fb0d4164996c93) should resolve this issue by removing them from the persistent data path and adjusting the README to point users to the correct path that is currently working.

game_eggs/steamcmd_servers/v_rising/egg-v-rising.json
lilkingjr1: Commit [bd77d26](https://github.com/parkervcp/eggs/pull/1680/commits/bd77d26939a8b18887771d0d52fb0d4164996c93) should resolve this issue. I was mistaken that `StandardPvP.json` was identical to the default `ServerGameSettings.json` file, so the default value has been changed to `null`, which should load the default `ServerGameSettings.json` file (unless a custom `ServerGameSettings.json` file is created in `~/save-data/Settings`). A custom `ServerGameSettings.json` file is not created during the installation script in case users would like to use a game settings preset (a `ServerGameSettings.json` file being present would override the preset setting).

AeonLucid:
Thanks for the egg. I'm currently running it myself but I have a few questions.
1. Are the debug messages in the console supposed to be enabled by default?
2. The resulting docker container is 2.89 GB, by far my biggest one. Is this expected?

lilkingjr1:
@AeonLucid 
1. From what I understand, yes. [The documentation](https://github.com/StunlockStudios/vrising-dedicated-server-instructions#configuring-the-server) does not cover these settings' defaults, but the default `ServerHostSettings.json` file shipped with the server files in `~/VRisingServer_Data/StreamingAssets/Settings` does have it set to `true`, albeit it being limited to admins only as well. Because it can be limited to "admins" only, I presume it doesn't refer to debug messages printed to the console, but instead to players (or just admins by default) in game. If it still prints to you as an admin in game with the debug setting off, it may be related to a different issue regarding String Booleans in the JSON, but I doubt it.
2. If you are referring to the Docker ***container***, the README for the egg states that the minimum storage required for the server is 3.5 GB (it's relatively large due to it being a Unity project, where, by default, all of it's libraries and basic assets get shipped with the server files). If you are referring to the Docker ***image***, it is using the Wine Yolk by default, which according to it's [manifest](https://github.com/users/parkervcp/packages/container/yolks/23744152?tag=wine_latest) should be about 1 GB in size.

AeonLucid:
There is an issue with this egg. When you run this server on windows like normal and press CTRL+C, it properly saves and stops the server. This creates a new `AutoSave_X` directory.

> SaveOnExit() - Saving the world on exit!
UnityEngine.Logger:Log(LogType, Object)
UnityEngine.Debug:Log(Object)
ProjectM.GameBootstrap:SaveOnExit()
ProjectM.GameBootstrap:SaveAndShutdownServer()

When you stop the server in pterodactyl, the server is force closed and no world save is created. Meaning progress of the last _configured auto save seconds_ is lost.

lilkingjr1:
@AeonLucid Server process runs within Wine, so I'm not sure how to properly send a shutdown command to it. Currently, the egg sends a `SIGINT` to the Wine process. Not sure if Wine passes this along or not since I'm not an expert at Wine. If anyone knows how to do some `trap` magic or something, please let me know.

lilkingjr1:
@jwoodhouse I cannot replicate this on my machine, so it may possibly be something local to you.

SrLicht:
Even though the OS is in a different time zone the Docker still uses UTC, and for weekend events this leaves too much difference between the times.
image

The timezone of Pterodactyl and Wings is 
`` timezone: America/Argentina/Buenos_Aires``
`` APP_TIMEZONE=America/Argentina/Buenos_Aires``

parkervcp (approved):

PR #1730, opened 2022-06-29
[OpenRCT2] Fix Release Installations & Upgrade to PTDL_v2

lilkingjr1:
- Updated egg to now allow successful installation and running of release versions.
- Updated README with up to date information.

### All Submissions:

### DIFF
--- a/game_eggs/tycoon_games/openrct2/README.md
+++ b/game_eggs/tycoon_games/openrct2/README.md
@@ -50,9 +50,9 @@ ___
 
 |  | Minimum |
 |---------|---------|
-| Processor | Multiarch is supported for most modern processors  |
+| Processor | x86/64 (\*multiarch may be supported for develop builds, but has not been tested) |
 | RAM | 256-512 MiB (depends on # of clients & park size) |
-| Storage | 80 MB (*may require more if building from source*) |
+| Storage | 110 MB (*may require more if building from source*) |
 | RCT2 Files | **Not required** |
 
 ___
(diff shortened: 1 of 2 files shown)


Software-Noob:
Is there a particular reason for using a different `temp` folder, when `/tmp` exists, and is a specific tmp volume from wings? Space doesn't appear to be an issue, since I managed to (re)install multiple times.

Not a deal-breaker and otherwise LGTM, but I'm curious (same as appimages in use, that's new heh).

You are also deleting it at the beginning and the end, so a reinstall would trigger unnecessary errors, and deleting anything from `/tmp` is not required, because, it's `/tmp`.

lilkingjr1:
@Software-Noob Good questions.

The `temp` folder is created at `/mnt/server/temp` (on the server volume) as a staging area for downloading server files, potentially building them from source if using a develop build, and then moving needed files to different specific folders for their final location.

When I was originally developing the egg, I actually did try to download and build within the `/tmp` folder (which is on the installation volume). I honestly can't remember exactly what error I got, but I either got a build error when building from source or an out-of-storage error (or maybe it was a build error because it was out of storage?). Either way, I assumed installation volumes were prohibitively small because they are temporary and typically not used to download server files directly to them, so I moved the whole process to the server volume, which worked at the time. Maybe that isn't the case now? I just read the docs and they say the installation container *should* inherit the server's specified resource limits, which I would assume includes storage, but maybe not?

TL;DR: If `/tmp` works for both release ***and*** develop builds, and it's not dependent on how people have their Wings configured for installation containers (causing hard-to-diagnose fringe cases), then I would say using `/tmp` is the better way to go. I just haven't tested it again recently.
___
> You are also deleting it at the beginning and the end, so a reinstall would trigger unnecessary errors, and deleting anything from /tmp is not required, because, it's /tmp.

Not sure if you're confusing `/tmp` over my `/mnt/server/temp` here. I delete `/mnt/server/temp` at the beginning to make sure it doesn't already exist before building (in the case of a reinstall), and delete it at the end so that it gets cleaned up upon a successful build because the build files are no longer needed. If `/tmp` works, this would no longer be needed. It also shouldn't produce errors on a reinstall because it checks if the directory exists first: `[ -d temp ] && rm -rf temp`
___
> same as appimages in use, that's new

OpenRCT2's releases are... tricky. Their x86_64 zipped `.tar.gz` release build contains an executable that relies on a multitude of libraries/packages to run that are not currently in the Debian Yolk, nor are they packaged as `.so`'s within the zip. I also didn't want to bloat the Debian Yolk with all these packages just to get OpenRCT2 to work. Their `.AppImage` release build on the other hand... While it can't run directly (because apparently Debian Slim doesn't have FUSE integrated), it *does* have all the library `.so`s required to run in it, which can be easily extracted and used with `LD_LIBRARY_PATH`.

parkervcp (approved):

PR #1733, opened 2022-07-01
fix(jts3servermod): use yolks images and clean install script

QuintenQVD0:
### All Submissions:

The current version was from 2019 , I changed to the newer image and fixt the install script
Yes , It works

### DIFF
--- a/bots/teamspeak3/jts3servermod/egg-j-t-s3-server-mod.json
+++ b/bots/teamspeak3/jts3servermod/egg-j-t-s3-server-mod.json
@@ -4,7 +4,7 @@
         "version": "PTDL_v2",
         "update_url": null
     },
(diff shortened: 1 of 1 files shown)


Software-Noob:
What exactly does this fix in the install script?

QuintenQVD0:
> What exactly does this fix in the install script?

in past releases there was a sub folder in the zip. So now it just unzips directly in to /mnt/server because the move out of /tmp is not needed anymore because there is no sub folder in the .zip


before
-file.zip
--game
--files
    
Now:
file.zip
-files

Software-Noob (approved):
LGTM. Thank you for contributing :+1:

I've cleaned up the egg a bit as seen and explained in 3d73a8c5c3b27419d0c0b5a9f8442dd566ebec97

PR #1753, opened 2022-07-10
Add additional mv command for server jar

bobbyrc:
When attempting to use the `FTB-modpacks.ch Server` egg to install a modded FTB Minecraft server I ran into errors during the installation because Petrodactyl couldn't find the server jar. Upon inspecting the installed files, I realized that the naming of the file for the server jar that Pterodactyl was expecting did not match the jar in my server files.

The egg expects a file matching `forge-*.jar` during installation and attempts to rename that file to `forge-server.jar`. Petrodactyl will try to run `forge-server.jar` when booting up the server.

In my files I noticed that the server jar was actually named `minecraft_server.1.18.2.jar`. The modpack I'm trying to install is the "Plexiglass Mountain" pack, with a pack ID of 96. I tried both versions `1.0.2` and `1.1.1` and they both had the same naming scheme for the server jar.

To fix this, I've added an extra `mv` command to the installation script that should rename any jars beginning with `minecraft_server` to the correct name to be launched. I did not remove the `mv` command for jars matching `forge-*.jar`, and those jars will take precedence in the event there jars matching both cases.

I'm not sure if FTB has change the naming scheme of their server jars, or maybe it this comes down to OS differences. I'm running Ubuntu Server 18.04.6 LTS.

Thanks for taking a look at this!

### All Submissions:

  * If not, why?:


bobbyrc:
Actually in looking at this more it seems like this may not be a good fix because when launching the server with the `minecraft_server.1.18.2.jar`, even though it appears to work, Minecraft sees the server as a vanilla server.

Still though, at least on my end, the egg doesn't work without some manual configuration after downloading the server files.

Software-Noob:
Thanks for your contribution. Unfortunately, this does not fix any of the issues with Forge. You're renaming vanilla Minecraft jar. 

The Forge 1.17+ doesn't ship a single bundled jar anymore and instead uses JPMS. It would require a different approach similar to what was applied to the Forge egg using UNIX arguments. Although, it depends on how each individual modpack handles it as well.

PR #1781, opened 2022-07-28
Update pocketmine to work on ARM64

QuintenQVD0:
### All Submissions:

- move to pterodactyl installer image as it is build for  both arm64 and amd64
- add the ARCH variable: 
afbeelding
- add a bash if then , else function to check if the variable ARCH=amd64 if so just run the normal php install commands that where used in the original egg. else use the compile script from pmmp self to compile php for arm:
afbeelding

It installs fine on Oracle ARM64 and notthing has changed on the AMD64 install proces


kry008 (approved):
Tested on Oracle ARM

parkervcp (approved):

PR #1791, opened 2022-08-02
Starbound: use Ubuntu 22.04 + Workshop IDs

unquietwiki:
### All Submissions:

* [ X] Have you followed the guidelines in our Contributing document?
* [ X] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
  * If not, why?:

Workshop item loading via subscription wasn't working. Having it iterate through a list of subscribed items was more successful.

2. [X ] Have you tested your Egg changes?

### DIFF
--- a/game_eggs/steamcmd_servers/starbound/egg-starbound.json
+++ b/game_eggs/steamcmd_servers/starbound/egg-starbound.json
@@ -1,32 +1,30 @@
 {
     "_comment": "DO NOT EDIT: FILE GENERATED AUTOMATICALLY BY PTERODACTYL PANEL - PTERODACTYL.IO",
     "meta": {
-        "version": "PTDL_v1",
+        "version": "PTDL_v2",
         "update_url": null
     },
-    "exported_at": "2021-07-02T04:25:07+03:00",
+    "exported_at": "2022-08-02T08:05:23+02:00",
     "name": "Starbound",
     "author": "parker@parkervcp.com",
     "description": "Starbound takes place in a two-dimensional, procedurally generated universe which the player is able to explore in order to obtain new weapons, armor, and items, and to visit towns and villages inhabited by various intelligent lifeforms.",
-    "features": [
-        "steam_disk_space"
-    ],
-    "images": [
-        "quay.io\/parkervcp\/pterodactyl-images:ubuntu_source"
-    ],
+    "features": null,
+    "docker_images": {
+        "quay.io\/parkervcp\/pterodactyl-images:ubuntu_source": "quay.io\/parkervcp\/pterodactyl-images:ubuntu_source"
+    },
     "file_denylist": [],
     "startup": ".\/starbound_server",
     "config": {
         "files": "{\r\n    \"storage\/starbound_server.config\": {\r\n        \"parser\": \"json\",\r\n        \"find\": {\r\n            \"gameServerPort\": \"{{server.build.default.port}}\"\r\n        }\r\n    }\r\n}",
-        "startup": "{\r\n    \"done\": \"Starting UniverseServer\"\r\n}",
-        "logs": "{}",
+        "startup": "{\r\n    \"done\": \"Starting UniverseServer\",\r\n    \"userInteraction\": []\r\n}",
+        "logs": "{\r\n    \"custom\": true,\r\n    \"location\": \"logs\/latest.log\"\r\n}",
         "stop": "^C"
     },
     "scripts": {
         "installation": {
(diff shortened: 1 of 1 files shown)


gOOvER (commented):
I would use the Debian 11 ghcr Source Image

Software-Noob (changes requested):
Please update your egg to match with upstream before making changes, as you're adding deprecated logs and userInteraction back and removing egg features. Not good.

Also, use the [Yolks images](https://github.com/parkervcp/yolks), which include the Debian installer image with necessary packages and used for most of our eggs, thus speeding up the process and doesn't need yet another image on the system.

You're also making the MOD_LIST variable not viewable to the end user for some reason.

Wazbat:
We tried using this edited egg, however with the logs and startup settings reverted to how they are, and it at least starts to download properly now.

However, we're now getting the following errors:
```
Please use force_install_dir before logon!
Update state (0x41) staging, progress: 100.00 (1126979394 / 1126979394)
 Update state (0x81) verifying update, progress: 35.19 (396542586 / 1126979394)
 Update state (0x81) verifying update, progress: 72.11 (812714845 / 1126979394)
Error! App '211820' state is 0x602 after update job.
```

The good news though, is we found out that the account we were using wasn't actually logging in correctly, as we were using the email login codes instead of the mobile 2FA codes. It now works on the old image if the account has 2FA

gOOvER:
> We tried using this edited egg, however with the logs and startup settings reverted to how they are, and it appears to work.
> 
> However, we're now getting the following errors using the egg in this PR:
> 
> ```
> Please use force_install_dir before logon!
> Update state (0x41) staging, progress: 100.00 (1126979394 / 1126979394)
>  Update state (0x81) verifying update, progress: 35.19 (396542586 / 1126979394)
>  Update state (0x81) verifying update, progress: 72.11 (812714845 / 1126979394)
> Error! App '211820' state is 0x602 after update job.
> ```
> 
> The good news though, is we found out that the account we were using wasn't actually logging in correctly. It now works on the old image if the account logs in correctly (we belive the issue was using an account without the mobile 2fa, only the browser code)

This egg dont work. wouldn't use it. 

Your error  ` state is 0x602 ` is about your Internet connection and have nothing to do with the egg.

game_eggs/steamcmd_servers/starbound/egg-starbound.json
```diff
@@ -73,16 +75,18 @@
...
         },
         {
-            "name": "Use Workshop content",
-            "description": "Set to true if you want to use your subscribed Workshop content",
-            "env_variable": "WORKSHOP",
-            "default_value": "0",
-            "user_viewable": true,
+            "name": "Mod List",
+            "description": "Steam Workshop mod subscription list; single-spaced ID#s\r\n\r\nDefault mod to load: https:\/\/steamcommunity.com\/sharedfiles\/filedetails\/?id=1079381251 (Same spawn planet for all users)",
+            "env_variable": "MOD_LIST",
```
gOOvER, line 84: this Variable is nowhere used. Only variable in installscript is: ${WORKSHOP}

gOOvER (changes requested):

unquietwiki:
Sorry I haven't been active on this the past week; the guy that hosts Ptero for me has been having trouble on his end, so my Wing hasn't been active. I was trying to solve two problems with this; if that helps any...

1. Creating the container was an inconsistent process; and attempting to subscribe to mods, would not load correctly when the container was put to use (even some weird subdirectory issue).
2. In looking up how others were loading mods for their environments, I determined the non-subscription manual loading, was the better approach; which that new bash loop parses the numbers to implement. It was not my intention to obscure the mod list.

Thanks for looking into this.

Wazbat:
I noticed the bottom of the for loop appears to just create a symlink between the workshop folder and the mods folder

Does that work? Almost every guide a read for starbound said that it only reads `.pak` files in the root of the `mods` directory, not any in any of the child directories, so in this case the `.pak` files would still be in individual folders

unquietwiki:
@Wazbat that may not even be needed. The original setup was trying to dump the mods to the mod subdirectory, but there's still the one nested in the install. So I didn't want to break that link, in case it was there for a reason.

Wazbat:
> @Wazbat that may not even be needed. The original setup was trying to dump the mods to the mod subdirectory, but there's still the one nested in the install. So I didn't want to break that link, in case it was there for a reason.

Oh, I was under the impression that the old script did not leave the files nested? I'm pretty sure it created a link between the one pak file and created each link directly in the folder without nesting them

I have a PR open here too as there were some issues I was having with that part of the old script

gOOvER:
this PR can be closed, because fixed with #1829

PR #1819, opened 2022-08-14
Update magma egg

JustRed23:
# Description
Add support for the recently released 1.18 server jar. This also removes some unnecessary future proofing

  * If not, why?:

### DIFF
--- a/game_eggs/minecraft/java/magma/egg-magma.json
+++ b/game_eggs/minecraft/java/magma/egg-magma.json
@@ -1,10 +1,10 @@
 {
     "_comment": "DO NOT EDIT: FILE GENERATED AUTOMATICALLY BY PTERODACTYL PANEL - PTERODACTYL.IO",
     "meta": {
-        "version": "PTDL_v1",
+        "version": "PTDL_v2",
         "update_url": null
     },
-    "exported_at": "2022-05-22T00:20:47-04:00",
+    "exported_at": "2022-08-15T01:12:37+02:00",
     "name": "Magma",
     "author": "support@pterodactyl.io",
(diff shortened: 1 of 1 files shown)


JustRed23:
Woops accidentally uploaded the wrong file

JustRed23:
Should be good now. The only problem i have with this script is that the latest version behaves a little weird. It never checks if there is a server jar so when you use latest and the latest version is 1.18.2 it will try to run the installer jar method and fail. So you will have to set the version manually instead of setting it to latest

JustRed23:
I am not experienced with bash enough to fix this

Software-Noob:
> Should be good now. The only problem i have with this script is that the latest version behaves a little weird. It never checks if there is a server jar so when you use latest and the latest version is 1.18.2 it will try to run the installer jar method and fail. So you will have to set the version manually instead of setting it to latest

That's because you're checking if `MC_VERSION` is `latest` or not set to run the installer.

---

I've refactored a lot of the old code while reviewing it, but can't really finish as the installer hosts have outdated sources for required dependencies at the time of this comment. This makes testing the installer impossible.

What all versions currently have a custom bundled magma jar instead of relying on Forge UNIX args? According to my quick test and API results, all of them currently have a jar and don't need the installer. 

I'll try to run through all the jars later and drop the installer completely for current known versions if they indeed have a jar.

JustRed23:
Yea the current versions all support the server jar, the installer jar is kind of obsolete now.

QuintenQVD0:
This shoud already be fixt by: https://github.com/parkervcp/eggs/pull/1882

PR #1901, opened 2022-10-07
Update minio to successfully mark as running

JakeyPrime:
# Description

Guide is no longer the latest terminology in latest version of minio for the completion of startup.

### DIFF
--- a/storage/minio/egg-minio-s3.json
+++ b/storage/minio/egg-minio-s3.json
@@ -4,7 +4,7 @@
         "version": "PTDL_v2",
         "update_url": null
     },
-    "exported_at": "2022-07-08T16:16:10+02:00",
+    "exported_at": "2022-10-07T04:56:20-04:00",
     "name": "Minio S3",
     "author": "accounts@bofanodes.io",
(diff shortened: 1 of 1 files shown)


storage/minio/egg-minio-s3.json
```diff
@@ -16,7 +16,7 @@
     "startup": ".\/minio.sh",
     "config": {
         "files": "{}",
-        "startup": "{\r\n    \"done\": \"guide\"\r\n}",
+        "startup": "{\r\n    \"done\": \"guide\",\r\n    \"done\": \"Documentation:\"\r\n}",
```
lilkingjr1, line 19: Is having two "done" values valid? Does minio print one or other other depending on circumstances? If not, we should narrow it down to one that we're sure indicates that it started without error.

JakeyPrime: Yeah, it allows multiple. I have an egg with lots of different possible startup values (I think just over a dozen now) it's not pretty but it DOES work. 

Tested this egg myself today as I couldn't handle perpetual startup. 

I don't want people updating their egg but not their version of minio and having the same issue by removing the first rendition of startup, given this change wasn't long ago on Minio's part.

lilkingjr1: Awesome; thanks for the explanation. Backwards compatibility is always nice.

lilkingjr1 (approved):
LGTM

PR #1931, opened 2022-10-25
update terraria vanilla to allways pull latest

### LINKED ISSUE #1928: [Bug]: Terraria Vanilla
### Panel Version

1.10.4

### Wings Version

1.7.2

### Service

Terraria Vanilla

### Modified

No, I did not modify the egg

### Expected Behavior

Always get the latest version when reinstalling the server.

### Actual Behavior

Whenever a new version comes out, it never updates quickly, because when they release a new version they release it first on the website and no longer on the wiki which is where the egg gets. On the wiki they take a while to update with the new version, on the site whenever a new version comes out they also update there, so it's always quick to have the new vanilla version.

On the wiki itself they ask to fix the download link

image

At the bottom of the page: https://terraria.wiki.gg/wiki/Talk:Server

### Steps To Reproduce

Step 1 - Reinstall de Server
Step 2 - Gets the new server version
Step 3 - Still on previus version because the Wiki do not update yet

### Install logs

https://pteropaste.com/6effvd

QuintenQVD0:
# Description

Because bug fixes are only pushed to the main website and not the wiki where it currently downloads from.
I changed that if version is set to latest, it will use the website api to get the latest file zip name and then make the download url.
If version is set to anything else then latest, it will still pull from the wiki.

closes #1928 
And is tested with when a fix version was the latest and when a new releace came out. 
afbeelding

  * If not, why?:


parkervcp (changes requested):
In your curl use `-sSL` so it will follow and redirects for json file. 

This is more a safety thing so scripts won't fail

QuintenQVD0:
done: 
afbeelding

parkervcp (approved):
LGTM

PR #1975, opened 2022-11-24
added dropdown for map selection

dr-waterstorm:
# Description

Added dropdown variable for map selection in the startup configuration tab. Currently this includes the default map (Exiled Lands) and the DLC map Siptah.

  * If not, why?:

### DIFF
--- a/game_eggs/steamcmd_servers/conan_exiles/egg-conan-exiles.json
+++ b/game_eggs/steamcmd_servers/conan_exiles/egg-conan-exiles.json
@@ -4,7 +4,7 @@
         "version": "PTDL_v2",
         "update_url": null
     },
-    "exported_at": "2022-10-23T13:08:22+02:00",
+    "exported_at": "2022-11-24T09:47:43+01:00",
     "name": "Conan Exiles",
     "author": "brycea@terrahost.cloud",
     "description": "Conan Exiles is an open-world survival game set in the brutal lands of Conan the Barbarian. Survive in a savage world, build your kingdom, and dominate your enemies in brutal combat and epic warfare.",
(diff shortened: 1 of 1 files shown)


Software-Noob:
This will explicitly limit the end user to only these options without a way to use any other without update to the egg.

It's not a good solution to fields such as maps

dr-waterstorm:
I totally agree. However, the way Conan handles loading maps is not by using a name, but using the complete path. And sadly the path is not even in the same folders. So for example the default map location is: `/Game/Maps/ConanSandbox/ConanSandbox`
while the DLC is located in: 
`/Game/DLC_EXT/DLC_Siptah/Maps/DLC_Isle_of_Siptah`

Possible mods would be loacted in:
`/Game/Mods/ModPath/MapName`

The only other way I could think of would be to to allow the user to edit the full path. However, that does not work well with a clean interface in my opinion. 

Any suggestions? 
If you prefer to have the full path as a string input field I can change it. It's just more convenient for a user with the dropdown, but certainly does limit the user.

parkervcp:
> Conan handles loading maps is not by using a name, but using the complete path

A drop down makes the most sense only due to this fact. 

The issue is the that panel hosts would need to add maps for users to be able to select them.

> : Another suggestion would be to add a 3rd value "ConfigFile" (or similar), which does remove the startup parameter. In that case the user can setup the map via the config file.

While feasible, this is a technically challenging approach.

QuintenQVD0:
Wy not make a variable `MAP_LOCATION` and use that in the startup cmd and give some info about the default map path and the DLC map path and use that?

dr-waterstorm:
@parkervcp Yes, I concur.
@QuintenQVD0 Possible indeed, however I'm a fan of clean and easy interfaces. With this you'd need to put the path of all maps and a description in the text box of the input field. In my opinion this would clutter the interface and make it harder to read and use. But it's certainly a possibility.

Currently I'm using the first proposed version, but I'll happily change it to whatever is preferred in consent.

I'd suggest a dropdown with the two standard maps and a third option "custom" (not implemented yet), which simply removes all startup parameters. In that case a more advanced user can edit the config files and use any custom map there. This of course means that it will only load the map if the user puts the right things in the right config files, as Conan supports either startup params or config file options. This would satisfy the "ease of use" for standard users while still providing the complex options for users who know where to put what in the various config files.

Davidian9:
I'm interested in this request as well as I've made a custom egg with mod support. I had a user request support for the Savage Wilds map, and I ended up making the server map a drop down myself as @dr-waterstorm originally suggested.

image

I, however, also see the problem with this as we don't want to have to update the egg every time someone wants to use a new custom map. Outside of making this a well-documented **nullable|string** I don't really see another good way to do this. While achieving this via config files is possible, many of the recommendations I found really seem to prefer setting the map in the exec command. I plan to test my egg with the Age of War update this month, and then maybe try to share it as a second Mod supported egg similar to what is done for Unity game servers with BepInEx eggs.

With all that being said, I'll now go with a nullable string for the map Variable in my own egg. It also seems if someone prefers to use a drop down, it's rather trivial for them to edit the egg for their own server.

PR #2021, opened 2022-12-26
Fixed DayZ mod loading, removed quote marks around additional params

### LINKED ISSUE #1768: [Bug]: DayZ Experimental will not load mods
### Panel Version

1.9.2

### Wings Version

1.6.4

### Service

dayz experimental 

### Modified

No, I did not modify the egg

### Expected Behavior

adding mods to the startup tab, additional mods and server mods section, and having the mods load

### Actual Behavior

Mods do not load as the startup command that are automatically generated places some "" (quotation marks) in an incorrect place which doesn't allow the mods to load. The quotation that is being generated only incases the @mod, and editing the startup command to "-mod=@mod" has loaded the mod successfully.

Manually editing the startup command to add mods is a workaround.

### Steps To Reproduce

add mods to additional mods and server mods section of startup tab
start server
mods do not load

### Install logs

https://pteropaste.com/wcfz57/

Moondarker:
# Description
DayZ Experimental had a classic mistake in `-mod` and `-serverMod` launch options. Quotes must be placed before and after the *entire* argument, not just its value. See [ArmA 3 documentation](https://community.bistudio.com/wiki/Arma_3:_Startup_Parameters#Spaces) - while not documented for DayZ, the argument parser is the same for both games. If you need further proofs, see what DayZ Launcher itself passes to the game when launched with mods. Same goes for additional startup params, there's no sense in adding quotes right here as it will break once there's more than 1 additional parameter.

In it's current state, given multiple mods like `@Mod 1`, `@Mod 2` and `@Mod 3` as mod list, server will interpret these mods as `"@Mod 1`, `@Mod 2`, `@Mod 3` and `"`

Fixes #1768

  * ~~While I'm 146% confident this will work, I haven't tested it personally yet because I'm not the one who has this problem (I don't even have Pterodactyl installed yet), I merely fixed it for a user of my mod and looks like it worked for them. Will pull this PR out of draft state once I do test it.~~ Tested, it works and I was unable to escape arguments list and execute any commands - please try to do so in case I missed something tho.


Moondarker:
Note: this will break BattlEye RCON for those who had battleye config in server root... Should it be moved to `variables` or just rolled back to hard-coded `/`?

lilkingjr1:
> this will break BattlEye RCON for those who had battleye config in server root

How so? How does moving the quotation marks affect BattlEye? Or are you just saying it's broken in the first place?

> Should BEpath be moved to variables or just rolled back to hard-coded `/`?

So I just read the documentation and found something interesting, though I don't know if it's been updated since I made the egg or if there was a specific reason why I deviated from it. It states:

> By default BattlEye will create BattlEye folder inside server profile folder.

I'm assuming that's the `-profiles` folder. For `-profiles` it states:

> Does not work on Linux, no profile will be created or used; instead, remove this parameter and the profile will be created in `~/.local/share/Arma 3 - Other Profiles`.

This is funny, because specifying it *does* actually work, but that's besides the point. `~` *should* resolve to `/home/container/`, which means (if it works) we could just remove the `-profiles` and `-bepath` flags to be clean and default. My only concern at the time of writing the egg was to ensure both the profile folder and BE folder resided within `/home/container/` where the end-user could actually reach it if desired.

Moondarker:
> How so? How does moving the quotation marks affect BattlEye? Or are you just saying it's broken in the first place?

I meant that if somebody were to update the egg version on existing server, their BattlEye config wouldn't load, but that's not a concern anymore as I rolled that back in c328485

> So I just read the documentation and found something interesting, though I don't know if it's been updated since I made the egg or if there was a specific reason why I deviated from it. It states:
> 
> > By default BattlEye will create BattlEye folder inside server profile folder.
> 
> I'm assuming that's the `-profiles` folder. For `-profiles` it states:
> 
> > Does not work on Linux, no profile will be created or used; instead, remove this parameter and the profile will be created in `~/.local/share/Arma 3 - Other Profiles`.
> 
> This is funny, because specifying it _does_ actually work, but that's besides the point. `~` _should_ resolve to `/home/container/`, which means (if it works) we could just remove the `-profiles` and `-bepath` flags to be clean and default. My only concern at the time of writing the egg was to ensure both the profile folder and BE folder resided within `/home/container/` where the end-user could actually reach it if desired.

That's a typical BI documentation moment ;D

Moondarker:
Alright, I've managed to test everything, and it looks like quotes placement doesn't matter at all on Linux, unlike Windows

The reason why mods are not loading e.g. in #1768 seems to be [a decades old bug in DayZ itself](https://github.com/GameServerManagers/LinuxGSM/discussions/3899#discussioncomment-3817188) - it only works fine when filesystem is case-insensitive. As soon as you introduce an upper-case letter _anywhere_ past the mod name it breaks (e.g. `@CF\addons\gui.pbo` is fine, `@CF\Addons\GUI.pbo` is not)

PR #2022, opened 2022-12-26
Fixed Arma 3 mod loading, removed quote marks around additional params

Moondarker:
# Description
Arma 3 had a classic mistake in `-mod` and `-serverMod` launch options. Quotes must be placed before and after the *entire* argument, not just its value. See [ArmA 3 documentation](https://community.bistudio.com/wiki/Arma_3:_Startup_Parameters#Spaces). If you need further proofs, see what ArmA 3 Launcher itself passes to the game when launched with mods. Same goes for additional startup params, there's no sense in adding quotes right here as it will break once there's more than 1 additional parameter.

In it's current state, given multiple mods like `@Mod 1`, `@Mod 2` and `@Mod 3` as mod list, server will interpret these mods as `"@Mod 1`, `@Mod 2`, `@Mod 3` and `"`

Basically #2021 for ArmA 3

* [ ] Have you tested and reviewed your changes with confidence that everything works? 
  * While I'm 146% confident this will work, I haven't tested it personally yet (I don't even have Pterodactyl installed atm)

### DIFF
--- a/game_eggs/steamcmd_servers/arma/arma3/README.md
+++ b/game_eggs/steamcmd_servers/arma/arma3/README.md
@@ -34,6 +34,14 @@ ___
             <a href="https://github.com/parkervcp/eggs/commits?author=IAmSilK" title="Contributor">💡</a>
         </td>
         <td align="center">
+            <a href="https://github.com/Moondarker">
+                <img src="https://avatars.githubusercontent.com/u/4098364" width="50px;" alt=""/><br /><sub><b>Moondarker</b></sub>
+            </a>
+            <br />
+            <a href="https://github.com/parkervcp/eggs/commits?author=Moondarker" title="Codes">💻</a>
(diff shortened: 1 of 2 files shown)


lilkingjr1:
Thank you for the PR and link to documentation. While the existing egg works as-is, it is probably better to have things be proper according to documentation. I (or someone else) will have to test this (just in case) before it is merged, but my panel is also down at the moment sadly. Lastly, please return the escaped quotations back to the `STARTUP_PARMS` variable. While it may not be 100% proper, if forces Arma to interpret the string vs. bash, which protects against end-user exploits. Yes, the closing `"` may be read by Arma as an extra character to interpret, but it will simply just throw it away (error free) like it does now when multiple flags are added (ex. `-noLogs -limitFPS=60` works fine).

Moondarker:
Thanks for the fast reply! You're right, I'll try to come up with a better solution for startup params asap

PR #2075, opened 2023-01-20
Tower Unity - added deprecated note

gOOvER:
# Description

see title


QuintenQVD0:
Does the egg still work? If not then you can remove it.

gOOvER:
> Does the egg still work? If not then you can remove it.

it cant work, when dedicated servers are remove. They wanted to add them back 1 year ago. I got the egg working again, but it cant connect to their masterserverlist

QuintenQVD0 (approved):
THX!

PR #2077, opened 2023-01-22
Update egg-t-modloader.json

### LINKED ISSUE #2051: [Install Bug]: tModLoader
### Panel Version

1.10.3

### Wings Version

1.7.2

### Service

[terraria/tmodloader](https://github.com/parkervcp/eggs/tree/master/game_eggs/terraria/tmodloader)

### Modified

No, I did not modify the egg

### Expected Behavior

The version I selected cannot be installed, in this case `0.11.7.6`.

### Actual Behavior

The newest version is installed, due to an error in the installation's script - see screenshot:

chrome_smYLhsj5rj

Error: The older tModLoader versions were called different then than now!

### Steps To Reproduce

Select the version `0.11.7.6` or `v0.11.7.6` and try to install it, you will get the latest version.

### Install logs

https://ptero.co/evidedaliq.properties

kjj194:
By adding the ?per_page=100 query to the install script we can expand the amount of supported release versions to 100 up from the default of 30.

This will temporarily fix #2051 Though when a version drops off the first page of 100 releases an iterative process to check all results may need to be added. 

If anyone is looking to add this additional functionality the appropriate command would be ?page=x where x is the page of results to load.

  * If not, why?:


game_eggs/terraria/tmodloader/egg-t-modloader.json
```diff
@@ -22,7 +22,7 @@
...
...n\r\n    tar xzvf ${DOWNLOAD_LINK##*\/}\r\nelif [ \"$FILETYPE\" == \"Zip\" ]; then\r\n    unzip -o ${DOWNLOAD_LINK##*\/}\r\nelse\r\n    echo -e \"unknown filetype. Exiting\"\r\n    exit 2\r\nfi\r\n\r\nif [[ \"$VERSION\" == v0* ]]; then\r\n    chmod +x tModLoaderServer.bin.x86_64\r\n    chmod +x tModLoaderServer\r\nelse\r\n    #tiny startup script for backward compatibility\r\n    echo 'dotnet tModLoader.dll -server \"$@\"' > tModLoaderServer\r\n    chmod +x tModLoaderServer\r\nfi\r\n\r\necho -e \"Cleaning up extra files.\"\r\nrm -rf terraria-server-*.zip rm ${DOWNLOAD_LINK##*\/}\r\nif [[ \"$VERSION\" != v0* ]]; then\r\n    rm -rf DedicatedServerUtils LaunchUtils PlatformVariantLibs tModPorter RecentGitHubCommits.txt *.bat *.sh\r\nfi\r\n\r\n## using config for difficulty as the startup parameter does not work -> config parser\r\nmv \/mnt\/server\/serverconfig.txt \/mnt\/server\/config.txt\r\nsed 's\/#difficulty\/difficulty\/' \/mnt\/server\/config.txt > \/mnt\/server\/serverconfig.txt\r\nrm \/mnt\/server\/config.txt\r\n\r\n## install end\r\necho \"-----------------------------------------\"\r\necho \"Installation completed...\"\r\necho \"-----------------------------------------\"",
```
parkervcp, line 25: If you are going to edit an egg please do it in panel and export the egg.

parkervcp (changes requested):

game_eggs/terraria/tmodloader/egg-t-modloader.json
kjj194: Ah my bad, is this better?

QuintenQVD0:
Ehh. Thar release he wanted is almost 3 years old. Then tmodloader still used mono and not dontet. And the release file on git w1s named difrend so the match variable greped notthing. Just chaning that will liky still not work. You made a pr for a issue marked as not planed? Does the sever even install this way?

kjj194:
I made the pr mostly because I needed the older version of tmodloader for something I was working on. Noted that the api call was only getting the first 30 releases and then defaulting to latest if version didn't match one of those. Figured I'd put this out in case anyone else runs into the same issue. I only found the issue when ticking off the checklist for submitting a pr.

kjj194:
Also I tested all the way back to v0.11.8.5 and it was working. Not sure back beyond that.

QuintenQVD0:
how does that even works , it greps for tmodloader.zip
afbeelding

parkervcp:
This would fail to grab get the releases you say it should. I also see no reason to support a two and a half year old release that probably has no support on their side either.

If someone wants this they can maintain their own version for older releases.

PR #2095, opened 2023-01-27
Add: SFTP Storage Share

lilkingjr1:
# Description

Adds a simple "empty" egg that acts as a private SFTP storage share by utilizing Pterodactyl's built in SFTP system for servers.

## New egg Submissions


QuintenQVD0 (approved):

scharez:
Is it possible to limit the disk space? When i leave the server running it should work, am i right?

lilkingjr1:
> Is it possible to limit the disk space? When i leave the server running it should work, am i right?

@scharez Currently, the start up script just prints something and then the server shuts down. It may get rebooted a couple of times by Pterodactyl because it "crashed", but it will eventually stay off. Disk space is only checked if the server is running, and is a limitation of Pterodactyl. If you modify the startup command to run something on loop to keep the server alive, that may work.

scharez:
> > Is it possible to limit the disk space? When i leave the server running it should work, am i right?
> 
> @scharez Currently, the start up script just prints something and then the server shuts down. It may get rebooted a couple of times by Pterodactyl because it "crashed", but it will eventually stay off. Disk space is only checked if the server is running, and is a limitation of Pterodactyl. If you modify the startup command to run something on loop to keep the server alive, that may work.

Thanks for the reply. Do you have an idee which command i can define, to run something like ah loop?

QuintenQVD0:
> > > Is it possible to limit the disk space? When i leave the server running it should work, am i right?
> > 
> > 
> > @scharez Currently, the start up script just prints something and then the server shuts down. It may get rebooted a couple of times by Pterodactyl because it "crashed", but it will eventually stay off. Disk space is only checked if the server is running, and is a limitation of Pterodactyl. If you modify the startup command to run something on loop to keep the server alive, that may work.
> 
> Thanks for the reply. Do you have an idea which command i can define, to run something like a loop?

You can not. Even if you make a program that like checks disk space every 5 minutes on top of the wings build in one. It will shut the server and prevent it from starting. But the sftp will always keep working even of the server is ofline and the diskspace is full.

lilkingjr1:
@QuintenQVD0 He meant something that will keep the server running so Ptero will check disk space, but I don't remember if disk space is continuously checked by Ptero, or only on startup.

PR #2093, opened 2023-01-26
fix: .jar filename

kylhuk:
# Description

The *.jar file was renamed to e.g. `minecraft_server.1.18.2.jar` and this is why the server was not able to find the proper file to start the server after the installation. I'm not sure yet if this solves the issue completely, or if we have multiple, different `*.jar` filenames.
Old command: `mv forge-*.jar forge-server.jar`
New command: `mv minecraft_server.*.jar forge-server.jar`

Fixes #1731

## Checklist for all submissions

* [X] Have you followed the guidelines in our [Contributing document](https://github.com/parkervcp/eggs/blob/master/CONTRIBUTING.md)?
* [X] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
* [X] Have you tested and reviewed your changes with confidence that everything works?
* [X] Did you branch your changes and PR from that branch and not from your master branch?
  * If not, why?:

PR #2103, opened 2023-01-28
fix:  ftb modpacksch server installation script

kylhuk:
# Description

Previous discussion at #2093 

Fixed the installation script so FTB modpacks can be used again. Works for minecraft-pre-1.17 and -post-1.17.

  * If not, why?:


QuintenQVD0 (changes requested):
You always have to make the changes in the panel and export the egg and fully replace the egg not just what you want

parkervcp:
@kylhuk They are asking that you use the panel to export the egg.

QuintenQVD0:
> @QuintenQVD0 I just modified the JSON so it works. Could you please tell me what exactly isn't working with my changes?

That is not allowed. You edit the install script on the imported egg, press export then you get a updated .json and you replace the whole .json here with the .json you jsut exported

QuintenQVD0:
What is so difficult about this? You must always update the whole .json file when you make a PR. and a startup cmd can not be a user editable script

parkervcp:
@kylhuk I didn't check if you had exported the egg from the panel. I checked the script changes first. That was my primary concern. Quintens concerns are still valid as it guarantees that the format is correct.

As for your fix. I see you are just updating the start script. That is fine, but the main start command for the server is staying as it is. Please make the start changes a separate function with a check the file exists before running the.

parkervcp:
@kylhuk I am sorry this was a negative experience. I realized that I didn't explain fully why I closed the other PR. In our discussion I thought I got the point across. As for using the start script as the start command, that is a security issue that I should have also called out a long time ago in the rules.

I am updating the CONTRIBUTING.md and the pull request template right now as a response of this. I do appreciate that you were looking for a way to handle the issue with the egg.

PR #2085, opened 2023-01-24
fix(pocketmine): New PHP binary download URL

QuintenQVD0:
# Description

- change the PHP download link as there preparing for pm5 so the switched the naming for there php builds

closes #2082

## Checklist for all submissions

* [x] Have you followed the guidelines in our [Contributing document](https://github.com/parkervcp/eggs/blob/master/CONTRIBUTING.md)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
* [x] Have you tested and reviewed your changes with confidence that everything works?
* [x] Did you branch your changes and PR from that branch and not from your master branch?
  * If not, why?:

PR #2109, opened 2023-01-30
update bin downloader

royaljacques:
Patch of a download problem, the bin is not correctly downloaded in the container, and therefore the server does not start.


QuintenQVD0:
> Patch of a download problem, the bin is not correctly downloaded in the container, and therefore the server does not start.

Your base egg is way out of date as you removed the arm64 install.

QuintenQVD0:
Always update your egg first. And the PHP Download url was already changed #2085

PR #2129, opened 2023-02-11
pocketmine php

QuintenQVD0:
# Description

This is a temp fix for the PocketmineMP egg as there php PocketmineMP url is offline.
**This should not be merged as it needs github auth to work**

What I have done: This install script now download the php bin from githib artifacts here: https://github.com/pmmp/php-build-scripts This uses a query to the github api url to get the download link. This all works fine, but to be able to download the files you need to be authenticated. So a github token with read only is needed.

Direct RAW url: https://raw.githubusercontent.com/parkervcp/eggs/pocketmine-php/game_eggs/minecraft/bedrock/pocketmine_mp/egg-pocketmine-m-p.json

  * If not, why?:


HBIDamian:
This PR is now outdated. 
Bins can now be found at https://github.com/pmmp/PHP-Binaries/releases/ 
API: https://api.github.com/repos/pmmp/PHP-Binaries/releases
DL Link from Releases page: https://github.com/pmmp/PHP-Binaries/releases/download/php-8.0-latest/PHP-Linux-x86_64-PM4.tar.gz

Also, `get.pmmp.io` does this all for you, especially since it gets updates faster than what happens here.

PR #2144, opened 2023-02-18
Update RedM and FiveM for Option to set Game Version

rhysmcqueen:
# Description
Hello All, this my first PR ever so hopefully I did it right. Would like to be able to add a variable to the fiveM and redM eggs to allow for users to set the game version. This helps to allow server to be using update games. (Casino update, gang update rather than the version from 2018 or something)

Please let me know if I did anything wrong.

  * If not, why?:


QuintenQVD0:
It is required for the egg to be exported and fully replaced.

QuintenQVD0:
@rhysmcqueen The egg does support installing older versions of the game? so wy is this needed?

QuintenQVD0:
and this variable should not be required as server will need to run if that variable is empty

QuintenQVD0:
This is not needed here and can be done in Txadmin as that mostly will be used by end users.

PR #2146, opened 2023-02-19
FiveM - add arm check

gOOvER:
# Description

Added s simple ARM Check, because, its possible to install FiveM on ARM and then install fails, with "file not found"

  * If not, why?:


QuintenQVD0:
This will not work. The install image is amd64 only. So it will just say exec format error. And not even run the script

gOOvER:
looks like, it wokrs. A lot of people trying to install FiveM on arm and there was never any exec error; only file not found

QuintenQVD0:
It will say not found on start. And then we ask for the install logs that just say exec format error

QuintenQVD0:
And if the want to host gta v on arm64 tell them to use ragemp if this: #2134 gets merged

gOOvER:
> And if the want to host gta v on arm64 tell them to use ragemp if this: #2134 gets merged

Rage is not FiveM.

Thats the best solution to tell people, that an egg is not ARM Compatible and so lesser support and discussions about arm, as long ptero dont check this

QuintenQVD0:
There is no point on adding this as this will not even run on install. It is already added to the readme. We can add it to the startup cmd. But not in the install script

gOOvER:
> There is no point on adding this as this will not even run on install. It is already added to the readme. We can add it to the startup cmd. But not in the install script

noone reads the readme; Startup is the wrong place; correct place is installscripot, because it the first thing what runs after setting up a server.

QuintenQVD0:
It will not even run. Wy do you not understand that, this docker image ghcr.io\/parkervcp\/installers:debian is only build for amd64 so it will pull that and then just exit as the image is not build for the system arch, IT will not execute anything of the install script

QuintenQVD0:
zlos you removed the Download Link variable

gOOvER:
i dont have time to discuss; sry. I tested iot and it works

QuintenQVD0:
@gOOvER :
afbeelding

PR #2148, opened 2023-02-19
remove FlameCord

QuintenQVD0:
# Description

remove FlameCord as it we recently switch to building it from source as there prebuild .jars were removed, but they removed the files in there git repo so it can no longer be build

https://github.com/2lstudios-mc/FlameCord/commit/b95e9e3915afc6fd937e0be7f4eb68d073c650d5


gOOvER:
They only sell it:

https://builtbybit.com/resources/flamecord-mitigate-bots-exploits.13492/https://builtbybit.com/resources/flamecord-mitigate-bots-exploits.13492/updates

QuintenQVD0:
> They only sell it:
> 
> https://builtbybit.com/resources/flamecord-mitigate-bots-exploits.13492/updates

There is no way to auth to download it. So then just use the bunggecord egg , skip the install script and upload the paid jar

parkervcp (approved):

linsaftw:
Now compilable again. Sorry for that.

PR #2155, opened 2023-02-22
Update egg-phantom-bot.json

Fuggschen:
# Description

Edited variables because PhantomBot now has a setup page and they are no longer needed. It is recommended to use the setup page after first install. Users on PB Discord are also complaining that their keys are overwritten by the panel so removing them as variables from the panel would fix that too.

Tested the egg on my own instance as an update and as a new egg. Both works fine for me but new bot maybe needs additional testing. I just checked if the setup panel shows up and writes data to the config.

  * If not, why?:

### DIFF
--- a/bots/twitch/phantombot/egg-phantom-bot.json
+++ b/bots/twitch/phantombot/egg-phantom-bot.json
@@ -4,7 +4,7 @@
         "version": "PTDL_v2",
         "update_url": null
     },
-    "exported_at": "2022-10-02T18:28:42+02:00",
+    "exported_at": "2023-02-22T12:23:24+01:00",
     "name": "PhantomBot",
     "author": "mail@wuffy.eu",
     "description": "PhantomBot is an actively developed open source interactive Twitch bot with a vibrant community that provides entertainment and moderation for your channel, allowing you to focus on what matters the most to you - your game and your viewers.",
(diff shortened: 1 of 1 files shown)


parkervcp:
@Fuggschen I feel like there should be a way to bypass the setup pages if you set the values before it starts.

Fuggschen:
> @Fuggschen I feel like there should be a way to bypass the setup pages if you set the values before it starts.

Sorry for late reply. Yeah I could skip it like you do if you just update the bot and have a prefilled botlogin.txt. I have to work something out in bash to setup the file only on install and do nothing with it if it exist and remove the configfile part so it wont be replaced on every start

QuintenQVD0:
> > @Fuggschen I feel like there should be a way to bypass the setup pages if you set the values before it starts.
> 
> Sorry for late reply. Yeah I could skip it like you do if you just update the bot and have a prefilled botlogin.txt. I have to work something out in bash to setup the file only on install and do nothing with it if it exist and remove the configfile part so it wont be replaced on every start

you can always just use bash to check if a file exists like:
```sh
FILE=/mnt/server/botlogin.txt
if [[ ! -f  $FILE]]
then
    echo "$FILE does not exist on your filesystem."
fi
```
I think you can figure out yourself on how to use this

Fuggschen:
> > > @Fuggschen I feel like there should be a way to bypass the setup pages if you set the values before it starts.
> > 
> > 
> > Sorry for late reply. Yeah I could skip it like you do if you just update the bot and have a prefilled botlogin.txt. I have to work something out in bash to setup the file only on install and do nothing with it if it exist and remove the configfile part so it wont be replaced on every start
> 
> you can always just use bash to check if a file exists like:
> 
> ```shell
> FILE=/mnt/server/botlogin.txt
> if [[ ! -f  $FILE]]
> then
>     echo "$FILE does not exist on your filesystem."
> fi
> ```
> 
> I think you can figure out yourself on how to use edit this to your needs

Yeah sure! But I also have to modify the file in order to setup the bot if the user passes the key on installation etc. I'll figure it out. There are plenty eggs where I can look at how to do this lol

Fuggschen:
Okay totally didn't saw that the oauth keys aren't set by the panel. Wouldn't make any sense to force the user to edit the file via the editor when he could do it like now with the panel. And seems that the bug with the keys was some user error.

Closed

PR #2162, opened 2023-02-26
Changed max members and token field

PilaScat:
# Description
Updated to 12 max members and 60 for token length

### DIFF
--- a/game_eggs/yuzu-emu/yuzu-emu.json
+++ b/game_eggs/yuzu-emu/yuzu-emu.json
@@ -75,7 +75,7 @@
             "default_value": "4",
             "user_viewable": true,
             "user_editable": true,
-            "rules": "required|int|max:4|min:1",
+            "rules": "required|int|max:12|min:1",
             "field_type": "text"
         },
         {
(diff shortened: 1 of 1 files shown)


QuintenQVD0 (changes requested):
You did not read the checklist! this is not exported from a panel

PilaScat:
It's the same, I don't know how to make it otherwise

PilaScat:
In theory it shouldn't be because of my changes since the server hasn't started yet and the original docker I'm already using has always worked with these settings

PilaScat:
Here the compiler error: (I'm on x86_64)
image

Z3r0byte:
Must admit that I didn't look too closely at which values should be allowed there since I don't use them, so thanks for the PR. It seems indeed that the yuzu team has made some changes to their source code in the meantime, making that warning pop up. However, compilation still succeeds as its only a warning (just ran a reinstall and that worked).

Maybe there is some way to suppress that warning, although they can of course also be useful for debugging if something goes wrong, so I am not sure if that is something we want to do.

QuintenQVD0 (approved):

QuintenQVD0:
LGTM!

PR #2172, opened 2023-03-02
Updated The Isle: Evrima to the latest version

enozii:
# Description

Updated to add new variables in the config and to fix RCON variable

  * If not, why?:


game_eggs/steamcmd_servers/the_isle/evrima/egg-the-isle.json
```diff
@@ -52,10 +52,10 @@
             "name": "RCON",
             "description": "true or false",
             "env_variable": "RCON",
-            "default_value": "false",
+            "default_value": "=false",
```
QuintenQVD0, line 55: this = can be done in the file parser not here

game_eggs/steamcmd_servers/the_isle/evrima/egg-the-isle.json
```diff
@@ -52,10 +52,10 @@
...
             "user_viewable": true,
             "user_editable": true,
-            "rules": "required|string",
+            "rules": "required|in:=true,=false",
```
QuintenQVD0, line 58: same here

game_eggs/steamcmd_servers/the_isle/evrima/egg-the-isle.json
```diff
@@ -97,6 +97,16 @@
...
             "field_type": "text"
+        },
+        {
+            "name": "Humans",
+            "description": "Enable or disable humans. true or false",
+            "env_variable": "HUMANS",
+            "default_value": "=false",
+            "user_viewable": true,
+            "user_editable": true,
+            "rules": "required|in:=true,=false",
```
QuintenQVD0, line 108: and here

game_eggs/steamcmd_servers/the_isle/evrima/egg-the-isle.json
```diff
@@ -15,14 +15,14 @@
...
+        "files": "{\r\n    \"\/TheIsle\/Saved\/Config\/LinuxServer\/Game.ini\": {\r\n        \"parser\": \"file\",\r\n        \"find\": {\r\n            \"ServerName\": \"ServerName={{server.build.env.SERVER_NAME}}\",\r\n            \"MaxPlayerCount\": \"MaxPlayerCount={{server.build.env.PLAYER_COUNT}}\",\r\n            \"MaxPlayers\": \"MaxPlayers={{server.build.env.PLAYER_COUNT}}\",\r\n            \"RconEnabled\": \"RconEnabled{{server.build.env.RCON}}\",\r\n            \"RconPort\": \"RconPort={{server.build.env.RCON_PORT}}\",\r\n            \"RconPassword\": \"RconPassword={{server.build.env.RCON_PASSWORD}}\",\r\n            \"QueuePort\": \"QueuePort={{server.build.env.QUEUE_PORT}}\",\r\n            \"discord\": \"discord={{server.build.env.DISCORD}}\",\r\n            \"bEnableHumans\": \"bEnableHumans{{server.build.env.HUMANS}}\"\r\n        }\r\n    }\r\n}",
```
QuintenQVD0, line 18: No need to remove the = from RconEnabled here and the same for bEnableHumans

QuintenQVD0 (changes requested):

game_eggs/steamcmd_servers/the_isle/evrima/egg-the-isle.json
enozii: The reason that I ended up doing it this way is because if I only use "in:true,false", I get an error everytime I try to use it. 
image

QuintenQVD0: > The reason that I ended up doing it this way is because if I only use "in:true,false", I get an error everytime I try to use it. image

What is the regex you set to that variable.

enozii: This is the error I'm getting in the console. 
```
{"errors":[{"code":"ValidationException","status":"422","detail":"The selected value is invalid.","meta":
{"source_field":"value","rule":"in"}}]}
```
I'm simply trying to use `required|in:false,true` as the input, but whenever I do that it changes to the button and doesn't seem to work. If i do `boolean` then it works, but then the value in the file is either 0 or 1 and that's no good.

QuintenQVD0: it should work

QuintenQVD0: it works if you force it to be a string, I cleaned the egg a little up `required|string|in:false,true`

QuintenQVD0 (approved):
LGTM! thank you!

enozii:
Is there any reason why you have added Auto Update, App id and Beta ID? 
the app ID/beta ID is used in the install script, but the auto update isn't really doing anything/isn't used anywhere and the server will update in the startup script regardless.

QuintenQVD0:
> Is there any reason why you have added Auto Update, App id and Beta ID? the app ID/beta ID is used in the install script, but the auto update isn't really doing anything/isn't used anywhere and the server will update in the startup script regardless.

The variables are used for auto update, that is handeld in the docker image

PR #2205, opened 2023-03-28
Update egg-core-keeper.json

Attackwave:
# Description

Over a few hours of three player play, we encountered some issues. No problems have been identified so far after the changes, but the server is still being monitored. Since the Dedicated Server is still in Early Access, problems can always arise.

**Updated startup script to (hopefully) fix the following memory issues:**
- XVFB Error: no temporary space available
- Errorlog message "killed session". Server crashes at that point.
- infinitely growing memory requirement

**Changed:**
- added parameter "-batchmode"
- grep only important details from the log for output on the console

  * If not, why?:


QuintenQVD0 (approved):
LGTM!

PR #2210, opened 2023-04-03
Update the velocity egg

Jcodeerd:
# Description

Allow the input of the server jar name to be 31 characters, as "velocity-3.2.0-SNAPSHOT-247.jar" is 31 characters, therefor making 20 characters not enough.

* [x ] Have you followed the guidelines in our [Contributing document](https://github.com/parkervcp/eggs/blob/master/CONTRIBUTING.md)?
  * If not, why?:

* [n/a ] You verify that the start command applied does not use a shell script
  * [n/a ] If some script is needed then it is part of a current yolk or a PR to add one
* [  ] The egg was exported from the panel

### DIFF
--- a/game_eggs/minecraft/proxy/java/velocity/egg-velocity.json
+++ b/game_eggs/minecraft/proxy/java/velocity/egg-velocity.json
@@ -1,10 +1,10 @@
 {
     "_comment": "DO NOT EDIT: FILE GENERATED AUTOMATICALLY BY PTERODACTYL PANEL - PTERODACTYL.IO",
     "meta": {
-        "version": "PTDL_v1",
+        "version": "PTDL_v2",
         "update_url": null
     },
-    "exported_at": "2021-12-02T16:47:41+00:00",
+    "exported_at": "2023-04-03T16:55:25+02:00",
     "name": "Velocity",
     "author": "parker@parkervcp.com",
     "description": "Velocity is a Minecraft server proxy with unparalleled server support, scalability, and flexibility.",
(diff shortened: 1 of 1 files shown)


QuintenQVD0:
All changes to an egg must be made in a panel and then exported!

QuintenQVD0:
\+ this is pritty useless as it does not use the downloaded name from there website

It downloads them from the download url and then saves them as that variable name is set velocity.jar
so it will just just run: `curl -o ${SERVER_JARFILE} ${DOWNLOAD_URL}`

Wy would  you even change that as if a new version comes out it will still say the old version
```bash
curl -sSL https://api.papermc.io/v2/projects/velocity/versions/3.2.0-SNAPSHOT/builds/247/downloads/velocity-3.2.0-SNAPSHOT-247.jar -o velocity.jar
```

where the velocity.jar is the variable so it is not needed to change this
if you want to change the regex of this variable then set it to string, with a max of 32 and add the regex that it must end in .jar

Those changes must be made in a panel and then exported
IF you don't want to to that then feel free to ask me to do it or close this PR

Jcodeerd:
I modified it because I used a name instead of just calling it Velocity.jar and figured other users might benefit from it aswell.
I appreciate your feedback, I'll make the changes and export the egg via the panel in a bit!

QuintenQVD0:
> I modified it because I used a name instead of just calling it Velocity.jar and figured other users might benefit from it aswell. I appreciate your feedback, I'll make the changes and export the egg via the panel in a bit!

Already thank you for your time, I am also making some litte changes to the install script, so if it is oké for you i will push a commit to this PR and then you can test it out

Jcodeerd:
I think it's better keep them separate,  but I can try out your changes
aswell

Jcodeerd:
LGTM!

QuintenQVD0 (approved):

PR #2214, opened 2023-04-06
Make python invocation work with bash exec

joelsgp:
Currently, when using this egg, the "Stop" button in the panel will absolutely not work. This is because the stop button sends a SIGTERM signal to the Bash process, and the Bash process will ignore the SIGTERM without propagating it to the Python process.

By adding `exec`, the Bash process will be completely replaced by the Python process, so the SIGTERM will reach the Python process and it will terminate normally.

P.S. I just had a situation where the panel registered the server as completely shut down but the server was still up and running. Please fix it

# Description

* [ ] Have you tested and reviewed your changes with confidence that everything works?
  * If not, why?:

### DIFF
--- a/generic/python/egg-python-generic.json
+++ b/generic/python/egg-python-generic.json
@@ -18,7 +18,7 @@
         "Python 2.7": "ghcr.io\/parkervcp\/yolks:python_2.7"
     },
     "file_denylist": [],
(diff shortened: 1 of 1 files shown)


joelsgp:
It would also be nice if the documentation had details about what signals are being sent, and to where, when I click the Stop button, but that's a matter for another repo

parkervcp:
This won't work as the entire command is wrapped in an eval as part of the image. 

I have a PR for wings to make it possible to use exec but that is still in progress.

joelsgp:
do you have a preffered solution in the meantime? thinking of swapping out the bash binary

QuintenQVD0:
> do you have a preffered solution in the meantime? thinking of swapping out the bash binary

You can always try to set the stop cmd to `^^C`

parkervcp:
> swapping out the bash binary

Swapping out bash wont fix it as the eval is in the start script

Issue #2023, opened 2022-12-26
[Bug]: tModLoader


falceso:
### Panel Version

Wisp v2.4.33

### Wings Version

Wisp v2.0.28

### Service

tModLoader

### Modified

Yes, I modified the egg (will provide details below)

### Expected Behavior

The difficulty variable works when setting the value to `1` for expert and an expert world is generated.

### Actual Behavior

The difficulty variable is not considered in the command-line parameters and should be set in the `serverconfig.txt` instead, similar to the Terraria Vanilla egg. Don't forget to add `-config serverconfig.txt` to the startup line for the file to be read.

### Steps To Reproduce

Step 0 - Install a mod on the client and server such as Calamity that requires Expert Mode on the world.
Step 1 - Install a tModLoader server using this egg and set the difficulty variable to 1 in the startup parameters.
Step 2 - Log into the world and see the following error from the Calamity Mod icon.
![https://i.imgur.com/jEEVVER.png](https://i.imgur.com/jEEVVER.png)

_(Fix)_
Step 1 - Turn off the server, and head into the File Manager to edit `serverconfig.txt`, change on line `#difficulty=1` to `difficulty=1`. You may delete `-difficulty {{DIFFICULTY}}` from the startup command.
Step 2 - Change or delete the current world used.
Step 3 - Log into the server again and hover over this icon again.
![https://i.imgur.com/9AbiCjm.png](https://i.imgur.com/9AbiCjm.png)

### Install logs

https://pteropaste.com/u2saws/

QuintenQVD0:
Does it work if you change it the variable so the startup changes. Then delete the world folder and then start the server

falceso:
If I'm understanding what you're asking, I've forcefully set the difficulty to 1 as shown below and delete the world.
```yml
./tModLoaderServer -ip 0.0.0.0 -port {{SERVER_PORT}} -maxplayers {{MAX_PLAYERS}} -difficulty 1 -password "{{SERVER_PASSWORD}}" -motd "{{MOTD}}" -lang {{LANGUAGE}} -world ~/saves/Worlds/{{WORLD_NAME}}.wld -worldname {{WORLD_NAME}} -autocreate {{WORLD_SIZE}} -savedirectory ~/ -tmlsavedirectory ~/saves -modpath ~/mods
```

This does not appear to work. I have also consulted a "Knowledgeable"  person in the tModLoader Discord and they claim
> no its not
the game doesnt read it at all

[...]

PR #2216, opened 2023-04-09
Tmodloader fix config not being loaded at startup

MarcoCoreDuo:
# Description

I noticed that I couldn't get my world to generate in expert difficulty and then I saw that #2040 addressed #2023 but forgot to add the "-config serverconfig.txt" parameter.
This fixes the issue of the difficulty not being applied when a world is generated.

  * If not, why?:


parkervcp (approved):
LGTM

PR #2226, opened 2023-04-14
added foldingathome egg

Fuggschen:
# Description

Added a Folding@Home egg for easy to run worker node to contribute to the project. Only downside is that we can't use the WebUI because it would be exposed unprotected to the web. But I enabled FAHControl access with password protection.
Anonymous and user-specific folding (with username and passkey) is possible.

Team defaults to 0 and user to Anonymous. Power is set to full because cpu limit will limit the usage regardless of system cores. All is set in the startup command because it's pain to work with the XML. It doesn't follow standard XML syntax.

  * If not, why?:

## New egg Submissions

### DIFF
--- a/README.md
+++ b/README.md
@@ -345,6 +345,10 @@ If you are reading this it looks like you are looking to add an egg to your serv
 
 * [Elasticsearch](/software/elasticsearch)
 
+### Folding@Home
+
+* [Folding@Home](/software/foldingathome)
+
 ### Gitea
 
 * [Gitea](/software/gitea)
--- a/software/foldingathome/README.md
+++ b/software/foldingathome/README.md
@@ -1,0 +1,27 @@
+# Folding@Home
+
+[Folding@Home](https://foldingathome.org/) is a distributed computing project aimed at simulating protein folding and other molecular dynamics.
+
+## Remote Control
+
+You can connect via FAHControl to your instance. WebUI is not supported because it wouldn't be password protected.
+
+## CPU Usage Control
+
+There are three power modes:
+- light
+- medium
+- full
+
+Only light and medium are useful because full enables gpu which is not supported in this egg.
+Light uses half of the available cpu cores **of your system**. Full uses **all of your cpu cores.** You can still limit the cpu usage via the panel.
+
+## Supported Versions
+
+Versions < v8 are supported. v8 is still beta and wont work with this egg. Get the latest v7 version number here: [Folding@Home Downloadserver](https://download.foldingathome.org/releases/public/release/fahclient/debian-stable-64bit/)
+
+## Server Ports
+
+| Port | default |
+| ---- | ------- |
+| FAHControl  | 36330    |
--- a/software/foldingathome/README.md
+++ b/software/foldingathome/README.md
@@ -2,6 +2,14 @@
 
 [Folding@Home](https://foldingathome.org/) is a distributed computing project aimed at simulating protein folding and other molecular dynamics.
 
+## User Account
+
+You can setup an User Account to track your PPD (Points per Day) and done WUs (Work Units). To get a Passkey visit the [Folding@home Passkey Site](https://apps.foldingathome.org/getpasskey)
+
+## Teams
+
+You can add your earned Points to a Team that you support. This is not nessesary as the points are just to track how much work was done (with some exeptions). Here is the [full List of the Teams](https://stats.foldingathome.org/team)
+
 ## Remote Control
 
 You can connect via FAHControl to your instance. WebUI is not supported because it wouldn't be password protected.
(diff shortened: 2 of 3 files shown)


skittles-fivem:
add in a max cpu limit of 64 threads allowed to be used in the config.xml or it will only use 1 thread

Fuggschen:
> add in a max cpu limit of 64 threads allowed to be used in the config.xml or it will only use 1 thread

@skittles-fivem can you tell me when this happens?
I never ran into this issue. My cores got all autodetected and utilisation goes up to 1200% on my 12 core CPU

QuintenQVD0:
does it respect the set cpu persentage? so 100% = 1 thread and 200% = 1 core
it can not just detect all resources and use everything. it can only use the allocated amount of cpu and ram

skittles-fivem:
> > add in a max cpu limit of 64 threads allowed to be used in the config.xml or it will only use 1 thread
> 
> @skittles-fivem can you tell me when this happens? I never ran into this issue. My cores got all autodetected and utilization goes up to 1200% on my 12 core CPU
im running 80 cores with Intel(R) Xeon(R) Gold 6138
by default i think the max is 64 anyway
ERROR:80 OpenMP threads were requested. Since the non-bonded force buffer reduction is prohibitively slow with more than 64 threads, we do not allow this. Use 64 or less OpenMP threads.

.
and yes it does respect the cpu % along with how much you set
gave it 64 cores with max of 5000% and it stayed at 5000%

Fuggschen:
> im running 80 cores with Intel(R) Xeon(R) Gold 6138 by default i think the max is 64 anyway ERROR:80 OpenMP threads were requested. Since the non-bonded force buffer reduction is prohibitively slow with more than 64 threads, we do not allow this. Use 64 or less OpenMP threads.
Do you have an example config.xml? I can't find a good documentation about setting a default max core setting. And is this necessary if it's already max 64 threads?

@QuintenQVD0 Yes! If you set your limits to 0 it will use everything. But you can just set a percentage and it won't exceed it. You could also pin it to core 0,1 and it would only use 200% cpu

skittles-fivem:
so on the cpus v is how many cores(threads) you want to apply to it and they dont like more then 64
<config>
  
  <slot id='0' type='CPU'/>
  <cpus v='64'/>
       
</config>

Fuggschen:
> so on the cpus v is how many cores(threads) you want to apply to it and they dont like more then 64 [config.txt](https://github.com/parkervcp/eggs/files/11238830/config.txt) but change it to .xml github didnt like xml

Okay thanks! Thougt this is to overwrite how many cores it HAS to use. Not what it could use. Works on my 12 core and still uses only 6 in light mode. Will add this shortly

skittles-fivem:
it auto detects it but it will attempt all cores unless changed to do so which in my case forces 1 thread since it doesnt know how many to apply

skittles-fivem:
could add in https://apps.foldingathome.org/getpasskey for the passkey help

Fuggschen:
> could add in https://apps.foldingathome.org/getpasskey for the passkey help

Updated. Also added Passkey and Teams to readme

skittles-fivem:
looks good to me

Fuggschen:
Had to fix my mail again. Panel don't want to save the changes in my mail address..

QuintenQVD0:
jus a question. Pterodactyl does support ARM64 and you have arm64 bin so do you mind adding it? if not then that is no problem then I just do it and add the commit to this. If you want to do it yourself that is 100% fine let me know if you need some help.

a hit to get you started: `ARCH=$([[ "$(uname -m)" == "x86_64" ]] && echo "amd64" || echo "arm64")` ( you ofc can change the egg to what you need / what is the easyest for you) is normaly our switching logic to download difrend binary's. 
Note that you will have to change the install image as it is AMD64 only

Fuggschen:
> jus a question. Pterodactyl does support ARM64 and you have arm64 bin so do you mind adding it? if not then that is no problem then I just do it and add the commit to this. If you want to do it yourself that is 100% fine let me know if you need some help.
> 
> a hit to get you started: `ARCH=$([[ "$(uname -m)" == "x86_64" ]] && echo "amd64" || echo "arm64")` ( you ofc can change the egg to what you need / what is the easyest for you) is normaly our switching logic to download difrend binary's. Note that you will have to change the install image as it is AMD64 only

Added it now. Needs testing because I don't have an ARM64 cpu but it should work just fine. It's just a change in the donwload link to 64bit or arm64

QuintenQVD0:
> > jus a question. Pterodactyl does support ARM64 and you have arm64 bin so do you mind adding it? if not then that is no problem then I just do it and add the commit to this. If you want to do it yourself that is 100% fine let me know if you need some help.
> > a hit to get you started: `ARCH=$([[ "$(uname -m)" == "x86_64" ]] && echo "amd64" || echo "arm64")` ( you ofc can change the egg to what you need / what is the easyest for you) is normaly our switching logic to download difrend binary's. Note that you will have to change the install image as it is AMD64 only
> 
> Added it now. Needs testing because I don't have an ARM64 cpu but it should work just fine. It's just a change in the donwload link to 64bit or arm64

I confirmed this works for amd64. I will check as soon as posible for arm64

QuintenQVD0 (approved):
LGTM! thank you

PR #2255, opened 2023-05-12
Update beamMP server egg

O1LER:
# Description

Added variables for server description and chat toggle
Updated variable "private" to accept only "true" or "false" string
Updated variable "max cars" to accept numbers up to 10

* [X ] Have you followed the guidelines in our [Contributing document](https://github.com/parkervcp/eggs/blob/master/CONTRIBUTING.md)?
* [ X] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
* [X ] Have you tested and reviewed your changes with confidence that everything works?
  * If not, why?:

* [X ] You verify that the start command applied does not use a shell script
* [ X] The egg was exported from the panel


QuintenQVD0:
you will need to edit the install script to when making a config file so it addes those lines by default

O1LER:
I see. I had no issue with it when testing the changes, since the beamMP server binary creates the serverConfig.toml itself upon first startup. So i guess the install script would not need to create the same file just prior. I will look into it, thanks for the hint

QuintenQVD0:
> I see. I had no issue with it when testing the changes, since the beamMP server binary creates the serverConfig.toml itself upon first startup. So i guess the install script would not need to create the same file just prior. I will look into it, thanks for the hint

You will have to manual add it to the install script as the configuration parser will prevent the server bin from writing its config as it is still editing it.

O1LER:
Updated the install script section by adding the LogChat variable
Corrected a typo (lower/upper case) in config file section, renamed chat logging (used to be called chat toggle)
Used same default description as beamMP

QuintenQVD0 (approved):
LGTM!

PR #2261, opened 2023-05-21
Update egg-subnautica.json

oliver408i:
# Description

Fixes issue with Subnautica 2.0 that downloads latest instead of legacy branch that Nitrox needs. When using this egg normally without the patch, nitrox will not run because the downloaded Subnautica version is 2.0, which Nitrox does not yet support.

  * If not, why?:

  * [ ] If some script is needed then it is part of a current yolk or a PR to add one (No new scripts)
* [ ] The egg was exported from the panel (it's a one-liner change in the install script only, which I copy pasted)


QuintenQVD0:
Not full exported for a panel. Until then this pr will not be revieuwed

oliver408i:
I added a commit with the exported json from the panel.

parkervcp (approved):

PR #2321, opened 2023-06-25
[Arma 3] Fixed mod parameter

maximilianovermeyer:
# Description

The mod parameter is defined as MODIFICATIONS, but in the startup command the non-existing parameter CLIENT_MODS was used.

I think [this one](https://github.com/parkervcp/eggs/pull/2022) drew a false conclusion why -mods isn't working.
  * If not, why?: I just branched quickly for this single change and won't keep the branch after a successful merge.

### DIFF
--- a/game_eggs/steamcmd_servers/arma/arma3/egg-arma3.json
+++ b/game_eggs/steamcmd_servers/arma/arma3/egg-arma3.json
@@ -14,7 +14,7 @@
         "ghcr.io\/parkervcp\/games:arma3"
     ],
     "file_denylist": [],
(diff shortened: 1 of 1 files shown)


QuintenQVD0:
You can not directly edit the .json file. It lust be done in the panel and exported fully. And I also think this is handled in the special images. 

@lilkingjr1 this is more for you.

lilkingjr1:
`CLIENT_MODS` is an internally rendered environment variable in the Yolk that is created during startup. Unfortunately, the downside is that it cannot be displayed properly in the startup command because it’s value is determined during startup, which can cause some confusion; but this is the only way to do it. In short, your proposed change would break the egg.

maximilianovermeyer:
That's quite surprising, mod ids added to MODIFICATIONS where ignored by our server, after changing the startup command, they worked. But then I will better switch to the mod file to avoid any unintended side effects.

lilkingjr1:
@maximilianovermeyer Works for me, so you must have an issue with formatting (read the variable description for instructions), mod folder names, or mod title casing. But yeh, modlist.html is definitely the easiest way if you have mod downloading enabled :)

PR #2324, opened 2023-06-26
removed unknown Disord ID and change it to a placeholder. No need to …

gOOvER:
removed unknown Disord ID and change it to a placeholder. No need to provide a unknown Discord ID, which could pose a safety risk

# Description

  * If not, why?:

### DIFF
--- a/bots/discord/ree6/config.yml
+++ b/bots/discord/ree6/config.yml
@@ -40,7 +40,7 @@ bot:
   misc:
     status: ree6.de | %guilds% Servers. (%shard%) # The Status of the Bot.
     feedbackChannelId: 0 # The Channel used for Feedback.
-    ownerId: <YOUR DISCOR ID> # The ID of the Bot Owner. Change this to yours!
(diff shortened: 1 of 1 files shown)


bots/discord/ree6/config.yml
```diff
@@ -40,7 +40,7 @@ bot:
   misc:
     status: ree6.de | %guilds% Servers. (%shard%) # The Status of the Bot.
     feedbackChannelId: 0 # The Channel used for Feedback.
-    ownerId: 321580743488831490 # The ID of the Bot Owner. Change this to yours!
+    ownerId: <YOUR DISCORD ID> # The ID of the Bot Owner. Change this to yours!
```
lilkingjr1, line 43: Good catch! Although, I would change it to `YOUR_DISCORD_ID` just to avoid using special characters in YAML (even though it probably still works as is).

parkervcp (approved):
Lgtm

PR #2332, opened 2023-06-30
Ark: Update startup to resolve issues with stop action

danny6167:
# Description

Corrects call to rcon tool. The -c flag shouldn't be there, that's to load a config file, not run a command

Calls processes async and then waits on them so that bash signal traps can actually work. STDIN is reattached to ensure input works still.
Traps *both* SIGINT and SIGTERM to account for wings bugs

  * If not, why?:


danny6167:
Issued as a draft PR right now as I think I can shorten the STDIN redirection path but need to test that first.

QuintenQVD0:
It visually looks fine to me. just don't have the current disk space to test it.

parkervcp (approved):

PR #2409, opened 2023-08-05
Update velocity egg

xXTheBigBearXx:
# Description

- Updates velocity.toml to the latest version, and changes any 127.0.0.1 entries to 172.18.0.1 (The default IP for pterodactyl0)

- Removes some text in the "Velocity Version Variable" referring to '1.1.x' as the latest dev version, as since Velocity merged into PaperMC the latest dev build is the same as the latest version

  * If not, why?:

### DIFF
--- a/game_eggs/minecraft/proxy/java/velocity/velocity.toml
+++ b/game_eggs/minecraft/proxy/java/velocity/velocity.toml
@@ -1,12 +1,12 @@
 # Config version. Do not change this
-config-version = "2.5"
+config-version = "2.6"
 
 # What port should the proxy be bound to? By default, we'll bind to all addresses on port 25577.
 bind = "0.0.0.0:25577"
 
 # What should be the MOTD? This gets displayed when the player adds your server to
-# their server list. Legacy color codes and JSON are accepted.
-motd = "&#09add3A Velocity Server"
+# their server list. Only MiniMessage format is accepted.
+motd = "<#09add3>A Velocity Server"
 
 # What should we display for the maximum number of players? (Velocity does not support a cap
 # on the number of players online.)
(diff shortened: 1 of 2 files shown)


parkervcp:
pterodactyl already replaces 127.0.0.1 with the ip address assigned to pterodactyl_nw. If anything this would break that and potentially break anyone who isn;'t using 172.18.0.1

game_eggs/minecraft/proxy/java/velocity/velocity.toml
```diff
@@ -72,25 +72,25 @@ enable-player-address-logging = true
 [servers]
 # Configure your servers here. Each key represents the server's name, and the value
 # represents the IP address of the server to connect to.
-lobby = "127.0.0.1:30066"
-factions = "127.0.0.1:30067"
-minigames = "127.0.0.1:30068"
+lobby = "172.18.0.1:30066"
```
parkervcp, line 75: This will break the panel updating the address on its own.

parkervcp (changes requested):

xXTheBigBearXx:
How does it perform that change?
From testing I could not connect to a server with a 127.0.0.1 IP assigned, with 127.0.0.1 in the .toml, but it would connect as soon as I changed it to 172.18.0.1 in the .toml

game_eggs/minecraft/proxy/java/velocity/velocity.toml
xXTheBigBearXx: Reverted to 127.0.0.1

parkervcp:
The panel updated the config on startup replacing the 127.0.0.1 in the toml on its own.

xXTheBigBearXx:
> on startup

Of the Container? If so it has never done so for me - is it something that needs enabling or should it be active by default?

parkervcp:
it doesn't touch the toml. I know it will update the server properties and the yaml for waterfall/bungeecord I believe.

QuintenQVD0:
You where right changing this to 172.18.0.1, the configuration parser does indeed not touch it.

parkervcp (approved):

PR #2449, opened 2023-08-31
Update egg-curse-forge-generic.json

dhkatz:
# Description

Fixed issue with CurseForge modpacks with escape characters in name.

Fixed issue with older versions of Minecraft in CurseForge modpacks.

  * If not, why?:


HerrSammyDE:
Was successfully tested by me

dhkatz:
May need to also look into my Modrinth egg to see if it needs to support ServerStarter too

rowbawts:
I'm currently testing this now and although it works it seems to use the JVM arguments for RAM specified in the `server-startup-config.yaml` and not the actual Ptero build configuration.

PR #2483, opened 2023-10-01
Specifies DB path in env variables

nateloeffel:
# Description
Fixes an issue where Uptime Kuma wouldn't be able to find the db correctly causing it to enter an infinite setup loop. By Specifying the correct directory an environmental variable it works correctly.

  * If not, why?:

  * [X] If some script is needed then it is part of a current yolk or a PR to add one

### DIFF
--- a/software/uptime-kuma/egg-uptime-kuma.json
+++ b/software/uptime-kuma/egg-uptime-kuma.json
@@ -77,6 +77,16 @@
             "user_editable": false,
             "rules": "nullable|string",
             "field_type": "text"
+        },
+        {
+            "name": "Data Directory",
+            "description": "The directory that Uptime Kuma will look for the kuma.db in.",
+            "env_variable": "DATA_DIR",
+            "default_value": ".\/db\/",
+            "user_viewable": false,
+            "user_editable": false,
+            "rules": "required|string|max:20",
+            "field_type": "text"
         }
     ]
 }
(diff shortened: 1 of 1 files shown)


gOOvER:
you didnt export the Egg.

for which Monitor is this needed?

nateloeffel:
This is for the Uptime Kuma egg, I just changed the environmental variable at the end. Would you prefer that I export the entire egg?

nateloeffel:
I along with many others got stuck in a setup loop. Try creating a new server with the egg that is currently on the repo and you will get stuck on a page not found/database setup loop.

gOOvER:
.db is the wrong dir. the the defaultpath is ./data

nateloeffel:
There must be some other issue then. Because when i create a server with this egg it just won't work period. It can't create the db-config.json at all and theres no way for it to run. Can you try and create a server with the egg and lmk if it works?

nateloeffel:
When I launch a new server using this egg, it gets stuck here:
image
If I try and go to the url then it displays this page
Screenshot 2023-10-01 at 1 12 12 AM

gOOvER:
Problem is not the env.

in some cases, the file `/data/db-config.json` will not created.

please create the file in path above with Content:

`{
    "type": "sqlite"
}
 `

and restar Container.

i will add a creation of this file in installscriot

gOOvER:
Found a better fix.  I'll do some tests and push a new egg :)

give me some time

QuintenQVD0:
In the currenr state this can not be merged. As it indeed will brake already running servers.

PR #2485, opened 2023-10-01
Update README.md

nateloeffel:
Important clarification that git config must be setup when the server is created and will not work correctly if setup afterwards.

# Description
Clarifies that you MUST setup the github credentials and the repository to pull from when you are creating the server, the feature will not work if you do it once the server is already installed.

  * If not, why?:

* [ ] You verify that the start command applied does not use a shell script

### DIFF
--- a/generic/nodejs/README.md
+++ b/generic/nodejs/README.md
@@ -20,3 +20,6 @@ You can use arrays to have multiple different values when different bots are bei
     "change this text 2"
   ]
 }
+```
+## Connecting to a Github Repository
+You can connect your server to automatically pull from a github repository, however the credentials for the repository must be provided and setup when the server is being installed.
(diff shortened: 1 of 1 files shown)


Software-Noob:
This is not entirely correct. The GitHub repository is cloned during the initial installation process. Therefore, if new credentials are provided afterward, a server reinstallation will be necessary

nateloeffel:
right, as opposed to just restarting the server

nateloeffel:
See if that looks better.

QuintenQVD0:
No one reads those README.md anyways. And there are a lot of more eggs (all generic) that use this script and logic.

nateloeffel:
It still appears confusing to someone who is using it for the first time. This way when they are confused why their server isn't cloning they can understand why.

QuintenQVD0:
> It still appears confusing to someone who is using it for the first time. This way when they are confused why their server isn't cloning they can understand why.

end users will never read this, only admins and they also wont so I don't find this needed

QuintenQVD0:
This is not realy needed, most uses know that and else they will come to the discord. There are a lot of eggs that use this logic and no one reads the README.md file anyway.

PR #2503, opened 2023-10-11
Update egg-7-days-to-die.json

Prisma501:
# Description

ControlPanelEnabled setting has been changed to WebDashboardEnabled in A21


QuintenQVD0:
You can not edit the json file directly. You must do it in a panel, export and fully replace the file here.

QuintenQVD0:
you could have just pushed a 2nd commit to this?

PR #2518, opened 2023-10-21
Update Sogbot

### LINKED ISSUE #2499: [Bug]: SogeBot ignoring VERSION variable
### Panel Version

1.11.3

### Wings Version

1.11.7

### Service

SogeBot

### Modified

No, I did not modify the egg

### Expected Behavior

Set a VERSION in variables and download and install that version.

### Actual Behavior

The Egg is isgnoring VERSION variable and still downloading and installing latest version from github.

### Steps To Reproduce

1. Set VERSION
2. Install
3. Run

### Install logs

https://pteropaste.com/bz8c610erb6d

QuintenQVD0:
# Description

- move to our official git release grab script
- modify the script if the release is not latest that the github api returns 50 results and not 25 if anyone is updating from a old version
- display the nodejs version as docker image and not the whole image
closes #2499 

  * If not, why?:

* x ] The egg was exported from the panel


gOOvER:
there is No Support for older Versions, only for latest. thats why the Script should not contain version

QuintenQVD0:
> there is No Support for older Versions, only for latest. thats why the Script should not contain versions
> 
> the used Version in #2499 was really old :) He didnt update for over 2 years :)

its limited to the latest 50 version so that is not that bad? if they have to upgrade it is posible, default is still latest

gOOvER:
in my eye's still not needed this change

QuintenQVD0:
The bot itself does not support older version. But we have a version variable so i find then that it should work or ir should just be removed.

PR #2526, opened 2023-10-26
Update egg-insurgency--sandstorm.json

Prophet731:
# Description

Replace the old image to one that's supported. This solves the problem of "GLIBCXX_3.4.26" missing.

ghcr.io/parkervcp/steamcmd:ubuntu

  * If not, why?:

### DIFF
--- a/game_eggs/steamcmd_servers/insurgency_sandstorm/egg-insurgency--sandstorm.json
+++ b/game_eggs/steamcmd_servers/insurgency_sandstorm/egg-insurgency--sandstorm.json
@@ -12,7 +12,7 @@
         "steam_disk_space"
     ],
     "images": [
-        "ghcr.io\/parkervcp\/games:source"
+        "ghcr.io\/parkervcp\/steamcmd:ubuntu"
     ],
     "file_denylist": [],
     "startup": "\/home\/container\/Insurgency\/Binaries\/Linux\/InsurgencyServer-Linux-Shipping {{MAP_NAME}}?Scenario={{SCENARIO}}?MaxPlayers={{MAX_PLAYERS}} -Port={{SERVER_PORT}} -QueryPort={{QUERY_PORT}} -log -hostname=\"{{HOSTNAME}}\"",


QuintenQVD0:
You can not change the .json file directly. Make the change in a panel and then export and fully replace the file fully.

QuintenQVD0:
The feature went a away. ANd you can not edit variables as the startup will not update of already made servers

Prophet731:
> The feature went a away. ANd you can not edit variables as the startup will not update of already made servers

What feature are you talking about?

Also this is what I had to do to allow sandstorm to allow things to work with game stats and such with the command lines.

Prophet731:
Added a fix for the hostname, switched to single quotes instead of double quotes. The shell was interpreting the `!` in the string causing problems.

QuintenQVD0:
I don't think you understand. You updated a variable, and the startup. But if someone updates this egg the variable names do update but the startup does not. So some things then will be empty.

Prophet731:
Fixed my mistake earlier now.

> I don't think you understand. You updated a variable, and the startup. But if someone updates this egg the variable names do update but the startup does not. So some things then will be empty.

The only thing that was removed was the "AUTO_UPDATE" but that isn't used in the startup config. The new variables I added are allowed to be empty so it shouldn't break any older versions.

QuintenQVD0:
> Fixed my mistake earlier now.
> 
> > I don't think you understand. You updated a variable, and the startup. But if someone updates this egg the variable names do update but the startup does not. So some things then will be empty.
> 
> The only thing that was removed was the "AUTO_UPDATE" but that isn't used in the startup config. The new variables I added are allowed to be empty so it shouldn't break any older versions.

Auto update also can not be changed. It is handled by the image internaly not in the startup.

game_eggs/steamcmd_servers/insurgency_sandstorm/egg-insurgency--sandstorm.json
```diff
@@ -24,8 +22,8 @@
...
-            "container": "ghcr.io\/parkervcp\/installers:debian",
+            "script": "#!\/bin\/bash\r\n## install required packages to install insurgency: sandstorm\r\napt update\r\napt -y --no-install-recommends install curl unzip libstdc++6 lib32gcc1 ca-certificates\r\n## install steamcmd\r\ncd \/tmp\r\ncurl -sSL -o steamcmd.tar.gz http:\/\/media.steampowered.com\/installer\/steamcmd_linux.tar.gz\r\nmkdir -p \/mnt\/server\/steam\r\ntar -xzvf steamcmd.tar.gz -C \/mnt\/server\/steam\r\ncd \/mnt\/server\/steam\r\n## needs to be used for steamcmd to operate correctly\r\nchown -R root:root \/mnt\r\nexport HOME=\/mnt\/server\r\n## install insurgency: sandstorm\r\n.\/steamcmd.sh +login anonymous +force_install_dir \/mnt\/server +app_update ${SRCDS_APPID} +quit\r\nmkdir -p \/mnt\/server\/.steam\/sdk32\r\ncp -v \/mnt\/server\/steam\/linux32\/steamclient.so \/mnt\/server\/.steam\/sdk32\/steamclient.so\r\nchmod +x \/mnt\/server\/Insurgency\/Binaries\/Linux\/InsurgencyServer-Linux-Shipping",
+            "container": "ubuntu:20.04",
```
parkervcp, line 26: why even change this?

parkervcp (dismissed):

game_eggs/steamcmd_servers/insurgency_sandstorm/egg-insurgency--sandstorm.json
gOOvER: becaus Debian 11 have only version 2.31-13+deb11u7 of glibc

QuintenQVD0: In the install container? Wy is that there even needed. Notthing runs there.

gOOvER: ok, i didnt follow the whole conversation. He changed ste steamcmd image to ubuntu. This should be enough, because its on 22.04

So the install image dont need a change.

But @Quinten; it would be better to update all images to Debian 12. i take a look into your PR, and as far i see, its ok. Only some packages need changes

Prophet731: > why even change this?

It's been a while since I even touched the install script side so I don't remember why that would be different.

gOOvER: > > why even change this?
> 
> It's been a while since I even touched the install script side so I don't remember why that would be different.

you can stay on the installer image. The game need glibc only, when the server is running and you allread chnge the steamcmd image to ubuntu, which is on 22.04

QuintenQVD0: > ok, i didnt follow the whole conversation. He changed ste steamcmd image to ubuntu. This should be enough, because its on 22.04
> 
> So the install image dont need a change.
> 
> But @Quinten; it would be better to update all images to Debian 12. i take a look into your PR, and as far i see, its ok. Only some packages need changes

You have me on discord. Some eggs will need updates to the install image like nodejs. Did not have time yet to make it.

QuintenQVD0 (approved):
LGTM! Thank you.

PR #2535, opened 2023-11-01
Fixes Zombied Server on Shutdown/Restart [ARK: Survival Ascended]

dagbs:
# Description

- `KeepAlive` has been added at shutdown due to extended up-times causing the next rcon command (`DoExit`) in-console to timeout; resulting in zombied servers.
- Added some very helpful troubleshooting tips to the README
- Added Known Issues to the README
- Bumped up the RAM requirements in the README

Reason for `KeepAlive` over `SaveWorld` is because `DoExit` does a `SaveWorld` and `KeepAlive` is not a command that's processed by the server. However, using `KeepAlive` will simply ping/pong the connection without causing an additional load as apposed to doing two `SaveWorld`(s) at the same time. Also, there is a slight chance that RCON will receive commands before the server is completely ready to do a proper `SaveWorld` where a `DoExit` will identify this properly and shutdown the server gracefully over a potential corrupted `SaveWorld`.

### DIFF
--- a/game_eggs/steamcmd_servers/ark_survival_ascended/README.md
+++ b/game_eggs/steamcmd_servers/ark_survival_ascended/README.md
@@ -5,14 +5,35 @@ ARK is reimagined from the ground-up into the next-generation of video game tech
 
 ## Troubleshooting
 
-Crash logs are available in the Files `ShooterGame/Saved/Crashes/UECC-Windows-*/CrashContext.runtime-xml`
+1) Try starting the server with no mods on a fresh install.
+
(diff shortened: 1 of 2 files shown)


dagbs:
This could all be resolved if the `rcon-cli` package also supported a retry flag that lets the command retry at least one more time on broken pipe or timeout.

gOOvER:
rcon-cli has no update sind roundabout 1 year

QuintenQVD0 (approved):
LGTM!

PR #1375, opened 2021-10-26
Add Satisfactory Egg

lilkingjr1:
# Add Satisfactory Egg
### Summary
This egg is a culmination of both @iamkubi 's and @matthewpi 's eggs that were posted on the Discord, but with a lot of clean up and polish. One major addition is the use of the new Pterodactyl Yolks images (specifically the `source` one), and also utilizes it's ability to update the server on start (if specified). Cosmetic additions like descriptions were also added, like port redirection clarification. Lastly, a freshly written README was included as well.

Egg has been fully tested on my machine, and clients are able to easily join. Additional testing from the community of this PR is welcome.
___

### All Submissions:

* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
(I know there's currently one #1374 open, but it looked like it might be closed due to thoroughness, and I wanted to keep this PR separate from that one)
* [x] Did you branch your changes and PR from that branch and not from your master branch?

### New Server Submissions:

1. [x] Does your submission pass tests (server is connectable)?
2. [ ] Does your server use a custom docker image?
    * [ ] Have you tried to use a generic image?
    * [ ] Did you PR the necessary changes to make it work?
3. [x] Have you added the server to the main README.md?
4. [x] Have you added a unique README.md for the server you are adding?

PR #2577, opened 2023-11-27
Satisfactory - Update ports to display the correct connection port

pedrotski:
Previously the port that is displayed to clients was the telemetry port, causing people to be confused as this isn't a port you can connect to with the default game.

# Description

This update changes the primary port to the server query port instead, so the correct connection port is displayed, and it moves the telemetry port to the back end. I've also updated the README to use the correct terminology.

  * If not, why?:

### DIFF
--- a/game_eggs/steamcmd_servers/satisfactory/egg-satisfactory.json
+++ b/game_eggs/steamcmd_servers/satisfactory/egg-satisfactory.json
@@ -4,7 +4,7 @@
         "version": "PTDL_v2",
         "update_url": null
     },
-    "exported_at": "2023-06-13T23:59:02+00:00",
+    "exported_at": "2023-11-27T01:35:48+01:00",
     "name": "Satisfactory",
     "author": "rehlmgaming@gmail.com",
     "description": "Satisfactory is a first-person open-world factory building game with a dash of exploration and combat. Play alone or with friends, explore an alien planet, create multi-story factories, and enter conveyor belt heaven!",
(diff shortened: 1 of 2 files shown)


game_eggs/steamcmd_servers/satisfactory/egg-satisfactory.json
```diff
@@ -15,7 +15,7 @@
         "ghcr.io\/parkervcp\/games:source": "ghcr.io\/parkervcp\/games:source"
     },
     "file_denylist": [],
-    "startup": ".\/Engine\/Binaries\/Linux\/*-Linux-Shipping FactoryGame ?listen -Port={{SERVER_PORT}} -ServerQueryPort={{QUERY_PORT}} -BeaconPort={{BEACON_PORT}} -multihome=0.0.0.0 $(if {{DISABLE_SEASONAL}}; then echo \"-DisableSeasonalEvents\"; fi)",
+    "startup": ".\/Engine\/Binaries\/Linux\/*-Linux-Shipping FactoryGame ?listen -Port={{TELE_PORT}} -ServerQueryPort={{SERVER_PORT}} -BeaconPort={{BEACON_PORT}} -multihome=0.0.0.0 $(if {{DISABLE_SEASONAL}}; then echo \"-DisableSeasonalEvents\"; fi)",
```
parkervcp, line 18: This change will break any already running server if they update to this egg...

parkervcp (changes requested):

game_eggs/steamcmd_servers/satisfactory/egg-satisfactory.json
pedrotski: I'm not sure if there is any way around this. It would be better to have the correct egg in the repo, though, rather than something that has been wrong from the start.

lilkingjr1: > I'm not sure if there is any way around this. It would be better to have the correct egg in the repo, though, rather than something that has been wrong from the start.

Please refer to these two links:
- [Egg README explaining ports](https://github.com/parkervcp/eggs/tree/master/game_eggs/steamcmd_servers/satisfactory#server-ports) (egg users are expected to read the README before using)
- [Original debate about ports](https://github.com/parkervcp/eggs/pull/1375#issuecomment-954991569) (this, and the two comments below it)

QuintenQVD0: To be honest I don't think this PR is needed, it is clearly mentioned in the readme and in the variable description that you have to join with the query port

pedrotski: Telling someone to read the docs because you've intentionally made something more difficult than it needs to be isn't going to make things better, though. It's just a bad experience for everyone who is involved.

QuintenQVD0: > Telling someone to read the docs because you've intentionally made something more difficult than it needs to be isn't going to make things better, though. It's just a bad experience for everyone who is involved.

that they not read the REAME that is so for most users, but not reading the variable description that is there own fault. This PR can not work and will brake all runing eggs as with an egg update the startup of already made servers will not upadate but the variables will.

QuintenQVD0: yes people do that, they just see there is an egg update, they update there egg and then see that there already made server does not work anymore and then they come to us in the discord.

parkervcp: If you are going to do anything just swap the query and server ports and don't rename it.

parkervcp: When you update an egg it overrides the egg on the old servers. Any server that already exists shouldn't update the start commands. That is a feature that hasn't been implemented yet.

pedrotski: > If you are going to do anything just swap the query and server ports and don't rename it.

I'll leave that for someone else to do as it's not correct.

parkervcp: updating the variable name is what breaks the currently running servers. You should just leave it named the way it is.

pedrotski: I understand, but its just supporting bad practices and will cause confusion with people. I appreciate your time.

Issue #2471, opened 2023-09-17
[Egg Request]: ntfy server 


diedyesterdaywashere:
### Does this expand an already existing service

No

### Link to a game or software

https://ntfy.sh

### Links for the download

https://github.com/binwiederhier/ntfy/releases/

### Links for the install docs

https://docs.ntfy.sh/install/

QuintenQVD0:
Pterodactyl is the wrong solution for this as you for like any action need to change the startup command and that is not secure

PR #2582, opened 2023-12-01
Add ntfy egg

ThnksCJ:
# Description

Ntfy is a simple HTTP-based [pub-sub](https://en.wikipedia.org/wiki/Publish%E2%80%93subscribe_pattern) notification service. It allows you to send notifications to your phone or desktop via scripts from any computer, and/or using a REST API.

Closed issue requesting ntfy: [#2471](https://github.com/parkervcp/eggs/issues/2471)

  * If not, why?:

* [ ] You verify that the start command applied does not use a shell script

## New egg Submissions


software/ntfy/egg-ntfy.json
```diff
@@ -0,0 +1,31 @@
...
+    "features": null,
+    "docker_images": {
+        "ghcr.io\/parkervcp\/yolks:debian": "ghcr.io\/parkervcp\/yolks:debian"
+    },
+    "file_denylist": [],
+    "startup": ".\/ntfy serve --config server\/server.yml",
+    "config": {
+        "files": "{\r\n    \"server\/server.yml\": {\r\n        \"parser\": \"yaml\",\r\n        \"find\": {\r\n            \"listen-http\": \":{{server.build.default.port}}\",\r\n            \"cache-file\": \".\/cache.db\"\r\n        }\r\n    }\r\n}",
+        "startup": "{\r\n    \"done\": \"INFO Listening on\"\r\n}",
+        "logs": "{}",
```
QuintenQVD0, line 20: Here should normaly ^c work aq tini is in that image.

software/ntfy/egg-ntfy.json
```diff
@@ -0,0 +1,31 @@
...
+        "stop": "^^C"
+    },
+    "scripts": {
+        "installation": {
+            "script": "## NTFY Installscript\r\n\r\n## update system\r\napt update\r\napt -y install --no-install-recommends curl tar\r\n\r\n## Variables\r\nARCH=$([[ \"$(uname -m)\" == \"x86_64\" ]] && echo \"amd64\" || echo \"arm64\")\r\n\r\ncd \/mnt\/server\r\n\r\n## install ntfy\r\necho -e \"download url: https:\/\/github.com\/binwiederhier\/ntfy\/releases\/download\/v2.8.0\/ntfy_2.8.0_linux_${ARCH}.tar.gz\"\r\necho -e \"downloading ntfy $VERSION\"\r\ncurl -sSL -o ntfy_2.8.0_linux_${ARCH}.tar.gz https:\/\/github.com\/binwiederhier\/ntfy\/releases\/download\/v2.8.0\/ntfy_2.8.0_linux_${ARCH}.tar.gz\r\n\r\necho -e \"unzipping ntfy exc\"\r\ntar -zxvf .\/ntfy_2.8.0_linux_${ARCH}.tar.gz ntfy\r\n\r\n## clear temp stuff\r\necho -e \"deleting temp files\"\r\nrm .\/ntfy_2.8.0_linux_${ARCH}.tar.gz\r\nrm -r .\/ntfy_2.8.0_linux_${ARCH}\r\n\r\necho -e \"creating cache.db\"\r\ntouch cache.db\r\n\r\necho -e \"-------------------------------------------------\"\r\necho -e \"Installation completed\"\r\necho -e \"-------------------------------------------------\"",
```
QuintenQVD0, line 25: Please use our github release grab script here. 

Some love to you for implementing arm64.

QuintenQVD0 (changes requested):

software/ntfy/egg-ntfy.json
ThnksCJ: Where can I find that script? Or do I just poke around in the eggs in the repo

ThnksCJ:
@QuintenQVD0 Ive added the changes you suggested could you look over it again?

QuintenQVD0:
looks fine to me, just have to test it.

QuintenQVD0:
@ThnksCJ How to we use this as the issue was closed because you like need to change the startup off the egg. so how does this egg handle this as I do not know this aplication

ThnksCJ:
You can configure the server.yml file that gets created and it's already pre-configured with a working config on the default allocation host and port. You can configure it to add a proxy host and such but you would need to look through the docs for that. If you span up an instance and visited it on http then it would all be working as a basic thing

ThnksCJ:
@QuintenQVD0 I changed the way the config is made to make it hopefully a bit more user friendly, could you look over it? Thanks 👍🏻

software/ntfy/egg-ntfy.json
```diff
@@ -15,14 +15,14 @@
     "file_denylist": [],
     "startup": ".\/ntfy serve --config server\/server.yml",
     "config": {
-        "files": "{\r\n    \"server\/server.yml\": {\r\n        \"parser\": \"yaml\",\r\n        \"find\": {\r\n            \"listen-http\": \":{{server.build.default.port}}\",\r\n            \"cache-file\": \".\/cache.db\"\r\n        }\r\n    }\r\n}",
```
QuintenQVD0: Wy did you remove this?

ThnksCJ: I removed this because this would only show when the server was started for the first time which I don't think is the best so I moved the server.yml creation into the init script.

QuintenQVD0: The files run on every startup and should be there as you must force the port it uses, and hard setting the cache db in it is also a good idea. The config just has to be in that place before the server starts for the first time.

ThnksCJ: Even though I create the file in the init script should I still keep it in the files? I just assumed it would be better as when the user installs the server they don't have to start it to see the configuration file

QuintenQVD0: End users do not see this configuration logic. it is expected that all ports an application uses are locked and users can not change it so they copy a config from somewhere, and it has another port set that it reverts that line back. else host will get flooded with support tickets.

Stef-00012:
how do you manage users if the startup command is hard coded?
(to manage users and permissions it's `ntfy user ...` and `ntfy access ...`

gOOvER:
> how do you manage users if the startup command is hard coded? (to manage users and permissions it's `ntfy user ...` and `ntfy access ...`

change the startup in the egg or server, when you are admin?

Stef-00012:
> > how do you manage users if the startup command is hard coded? (to manage users and permissions it's `ntfy user ...` and `ntfy access ...`
> 
> change the startup in the egg or server, when you are admin?

why? it would be enough to make the startup like `ntfy {{ARGS}}`
and make an env variable where the user can type `user add xxx` etc.

ThnksCJ:
I thought users could be added using the web UI and it work the same as using a cmd to do it

QuintenQVD0:
it should indeed, as the user will never be allowed to change the startup themselves in offical eggs.

Stef-00012:
> I thought users could be added using the web UI and it work the same as using a cmd to do it

you mean like so?
image

Stef-00012:
> I thought users could be added using the web UI and it work the same as using a cmd to do it

how? in the web UI the add user button is only a way to log into users made with `ntfy user add ...`, and in the web UI there's no way to manage permissions (`ntfy access ...`)

QuintenQVD0:
if that is true, this egg likely will be closed as uses are not allowed to self edit the startup

ThnksCJ:
> > I thought users could be added using the web UI and it work the same as using a cmd to do it
> 
> how? in the web UI the add user button is only a way to log into users made with `ntfy user add ...`, and in the web UI there's no way to manage permissions (`ntfy access ...`)

I must have overlooked this fact, mabye ntfy won't work as an egg after all

Stef-00012:
> if that is true, this egg likely will be closed as uses are not allowed to self edit the startup

can't it be done like so?
like `ntfy ${COMMAND}` 
where `COMMAND` is a `user_editable` variable

QuintenQVD0:
> > if that is true, this egg likely will be closed as uses are not allowed to self edit the startup
> 
> can't it be done like so? like `ntfy ${OPTIONS}` where `OPTIONS` is a `user_editable` variable

That would be better but is almost the same as just giving the user direct startup asses as you put a ; and the command and the panel will run it.

ThnksCJ:
Doesn't the nodejs egg do something like that to start the main file? Wouldn't you be able to put a ; and it escape too

QuintenQVD0:
> Doesn't the nodejs egg do something like that to start the main file? Wouldn't you be able to put a ; and it escape too

yes technically yes, but that variable is locked to a max of 64 char and should (it is not) be locked with alpha_dash rule what as it sas only allows letters and _ and not symbols.

ThnksCJ:
Then could you not do the same on this egg? Or would that raise password security issues

QuintenQVD0:
> Then could you not do the same on this egg? Or would that raise password security issues

Problem is it does not allow for `-` so many things (that's wy for node it is not set) will not work.
Even uppercase char are not allowed in alpha_dash. so I guess then indeed just lock it to 64 length and do it that way.

QuintenQVD0:
@ThnksCJ  Can an admin used create other users on the web interface or is it all command line interface, else we just create 1 stock user.

ThnksCJ:
From what i have read it all has to be done using the cli which is an issue, that is to create specific channels locked behind authentication. Not sure how we could overcome this issue without making a custom image that has a shell to accept commands or somehting

QuintenQVD0:
> From what i have read it all has to be done using the cli which is an issue, that is to create specific channels locked behind authentication. Not sure how we could overcome this issue without making a custom image that has a shell to accept commands or somehting

That wont be happening.

Stef-00012:
found out in the server.yml file you can add these 3 settings
```yml
enable-signup: true # anyone can create an account through the web ui
enable-login: true # anyone can login into accounts through the web ui using user and pass
enable-reservations: true # anyone can reserve topics for themselves
```

only limit is you can't choose who can make accounts and who can't, or who can reserve and who can't

however technically you can if you enable these settings only when you want to create an account or manage reserved topics and disable when you don't need to

QuintenQVD0:
> found out in the `server.yml` file you can add these 3 settings
> 
> ```yaml
> enable-signup: true # anyone can create an account through the web ui
> enable-login: true # anyone can login into accounts through the web ui using user and pass
> enable-reservations: true # anyone can reserve topics for themselves
> ```
> 
> only limit is you can't choose who can make accounts and who can't, or who can reserve and who can't
> 
> however technically you can if you enable these settings only when you want to create an account or manage reserved topics and disable when you don't need to

its even yaml so you could add variable for those nad let the config parser turn them on and off

PR #2607, opened 2023-12-14
Updated Evrima Egg

enozii:
Updated to work with the new map

# Description

* [x ] Have you followed the guidelines in our [Contributing document](https://github.com/parkervcp/eggs/blob/master/CONTRIBUTING.md)?
* [ x] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
* [ x] Have you tested and reviewed your changes with confidence that everything works?
  * If not, why?:

* [ x] You verify that the start command applied does not use a shell script
* [ x] The egg was exported from the panel


QuintenQVD0:
Is that the only option for MapName? if not then it should be done over a variable and the file parser

enozii:
It is the only map in the game yes, you'd have to change branch to play anything else. When they add multiple maps I'll make it a variable, but that's probably not gonna happen for a while given their dev time..

QuintenQVD0 (approved):
LGTM! thanks

PR #2608, opened 2023-12-15
Rathole

TunayAdaKaracan:
# Description

It is server side rathole egg for rathole software. 

I don't have much experience in creating an egg and overall linux. So i copied a lot from rust generic egg.
Note: It didn't work with debian yolk so i just instead used rust one which works fine.

  * If not, why?:

## New egg Submissions

### DIFF
--- a/software/rathole/README.md
+++ b/software/rathole/README.md
@@ -1,0 +1,7 @@
+# Rathole
+A secure, stable and high-performance reverse proxy for NAT traversal, written in Rust
+
+rathole, like frp and ngrok, can help to expose the service on the device behind the NAT to the Internet, via a server with a public IP.
+
+## This egg only includes server side of rathole
+Please refer to https://github.com/rapiz1/rathole for more information about rathole
--- a/software/rathole/egg-rathole.json
+++ b/software/rathole/egg-rathole.json
@@ -1,0 +1,31 @@
+{
+    "_comment": "DO NOT EDIT: FILE GENERATED AUTOMATICALLY BY PTERODACTYL PANEL - PTERODACTYL.IO",
+    "meta": {
+        "version": "PTDL_v2",
+        "update_url": null
+    },
+    "exported_at": "2023-12-15T15:44:13+03:00",
+    "name": "Rathole",
+    "author": "tunayada@gmail.com",
+    "description": "A rathole server egg designed to work with pterodactyl.",
+    "features": null,
+    "docker_images": {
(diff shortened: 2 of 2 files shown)


parkervcp:
I don't see why this would be needed for pterodactyl

TunayAdaKaracan:
I did for my self usage mainly as i have a raspberry pi. But wanted to share it. Rather it would be more useful if you are self hosting pterodactyl and not a daily user

QuintenQVD0:
This is a reverse proxy application, as web hosting is not supported and for other application will not work that well in docker containers, I am closing this

parkervcp:
They explained themselves and this can proxy more than just web services.

QuintenQVD0:
if you see it from that point then I guess this can stay open indeed

TunayAdaKaracan:
I would like to say, rathole is more of a tunneling software that can be alternative to ngrok and playit.gg. Even tho they start as free it is cheaper to afford a small ptero server if you are planning to tunnel more than one port. (Resource usage is really low with rathole). This could be useful for those who have single board computers or for those who don't have a static ip to port forward. It really depends on what you need to view this as useful or not. In my case, since i have a rasp pi i find it useful for tunnelling my socket applications, ssh and othee things.

TunayAdaKaracan:
One side note, unlike ngrok and playit. Your ip doesn't change with rathole. Imagine your wifi goes offline for sometime and you will have to login from home network to see what is your connection ip now. Doesn't sound fun, right?

QuintenQVD0:
> One side note, unlike ngrok and playit. Your ip doesn't change with rathole. Imagine your wifi goes offline for sometime and you will have to login from home network to see what is your connection ip now. Doesn't sound fun, right?

I have a pi myself, I just setup my pi to have a static reserved local ip and that is forwarded to the public ip what is also static

parkervcp:
> > One side note, unlike ngrok and playit. Your ip doesn't change with rathole. Imagine your wifi goes offline for sometime and you will have to login from home network to see what is your connection ip now. Doesn't sound fun, right?
> 
> I have a pi myself, I just setup my pi to have a static reserved local ip and that is forwarded to the public ip what is also static

This is calling out the external ip not being static, not the internal ip for the pi

TunayAdaKaracan:
Also, if you don't have a static ip, most likely you are in a CGNAT where you can't port forward

parkervcp:
> Also, if you don't have a static ip, most likely you are in a CGNAT where you can't port forward

This is not always true. I know Verizon in the US doesn't use CGNAT and you can use ddns to point at the address.

parkervcp (approved):

PR #2609, opened 2023-12-15
YuzuEmu - Add ARM support and update variables

MarcoCoreDuo:
# Description

1. The Yuzu Server binaries can be build on an ARM machine, but the debian install image doesn't support ARM, so I changed it to the official Bullseye image.
2. I removed the max length of the token field because my token is longer than 60 characters and there is no information about how long a token can be. I also changed the length of the Server name field to allow longer names.
3. Max Players recommended was raised to 16.
4. Lastly, I removed the slash at the end of the web-api url because it tried to fetch from https://api.yuzu-emu.org//... and this returned null.

I also renamed the egg to the panel exported file name.

  * If not, why?:


MarcoCoreDuo:
> Use `ghcr.io/parkervcp/installers:debian` it is for both amd64 and arm64

Done, thanks

QuintenQVD0 (approved):
LGMT! thanks.

PR #2610, opened 2023-12-16
Update Avorion

### LINKED ISSUE #2593: [Bug]: Avorion - Server name input gets cut off
### Panel Version

1.11

### Wings Version

1.11

### Service

avorion

### Modified

No, I did not modify the egg

### Expected Behavior

Server name in the build configurations should be written out correctly inside the server.ini file for the corresponding galaxy

### Actual Behavior

Server name gets partially written. It seems to only take the first word in a server name. If you place everything together such as

`--server-name MyPublicServer`

It works. But if it is

`--server-name My Public Server` it fails. Instead the ini file will only have the word "My" as the server name. Here are some screenshots as an example

Screenshot from 2023-12-06 23-02-54
Screenshot from 2023-12-06 23-03-21

### Steps To Reproduce

1. Import the egg to your avorion nest
2. For  "server name", input a name with white spaces and choose any name for the galaxy
3. Look at the corresponding galaxy's server.ini file after it is built or, look at the terminal output when it's finished.

### Install logs

https://pteropaste.com/pxcy6ra4giiu/

QuintenQVD0:
# Description

- change to the steamcmd debian image
- do the beta properly so also auto update will work
- add the auto update variable
- remove userInteraction
- fix that the server name can now have spaces
afbeelding
afbeelding

closes #2593 

  * If not, why?:

### DIFF
--- a/game_eggs/steamcmd_servers/avorion/README.md
+++ b/game_eggs/steamcmd_servers/avorion/README.md
@@ -22,3 +22,6 @@ Additionally the server can be configured with an RCON port.  RCON will be disab
 ## RCON
 
 If you plan to use RCON it currently needs to be manually configured in the server.ini file.  At some point server variables will be added.
+
+## Updating
+Because with an update to this egg, the startup command changed, so if you update this egg, you will manually have to update the startup command for every server that was already made.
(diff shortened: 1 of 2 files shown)


Voronsky:
This PR addressed the issue mentioned in #2593 , i was able to test it and it addressed the issues raised there.

parkervcp (approved):

PR #2633, opened 2023-12-29
Update minetest

Tealk:
# Description

Start parameters extended to correct the following error:
> 2023-12-29 18:22:36: ERROR[Main]: To run a Minetest server, you need to select a game using the '--gameid' argument.
> 2023-12-29 18:22:36: ERROR[Main]: Check out https://content.minetest.net for a selection of games to pick from and download.

Source:
https://wiki.minetest.net/Setting_up_a_server#Linux_2

* [] Did you branch your changes and PR from that branch and not from your master branch?
  * If not, why?: Because it is only a small change and I will delete the fork afterwards.


QuintenQVD0:
An egg must be exported from a panel!

QuintenQVD0:
and this should not be needed a it is auto set in the config 
default_game = {{server.build.env.DEFAULT_GAME}}

Tealk:
> An egg must be exported from a panel!

for 23 characters?


> and this should not be needed a it is auto set in the config default_game = {{server.build.env.DEFAULT_GAME}}

i just created a new server with the current egg and got this error message, with this fix the server could start.
otherwise the server would not start, if you do not like the fix you can also reject it and wait until users will open an issue about it.

QuintenQVD0:
> > An egg must be exported from a panel!
> 
> for 23 characters?
> 
> > and this should not be needed a it is auto set in the config default_game = {{server.build.env.DEFAULT_GAME}}
> 
> i just created a new server with the current egg and got this error message, with this fix the server could start. otherwise the server would not start, if you do not like the fix you can also reject it and wait until users will open an issue about it.
> 
> i have not actively followed the changes of minetest, but there have been many updates since the last egg update, maybe the switch is now absolutely necessary.

Yes for every change it must be done in a panel and exported.
If you export it then this can be merged as games do changes their server logic from time to time

QuintenQVD0 (approved):
LGTM! Thank you.

PR #2665, opened 2024-01-16
Update for new server binaries naming scheme

O1LER:
# Description

The beamMP egg needs to be updated in order to be able to pull the correct binary from https://github.com/BeamMP/BeamMP-Server/releases since the naming scheme got changed with v3.2.2 to accomodate a larger number of prebuilt binaries

  * If not, why?:


O1LER:
Removing the restriction on the MATCH variable is only the first step. Optimally, there would be checks that automatically determine if the host is x86 or arm, and if the distro is debian or ubuntu, since these options are now available prebuilt by beamMP. I will look into creating that

QuintenQVD0:
> Removing the restriction on the MATCH variable is only the first step. Optimally, there would be checks that automatically determine if the host is x86 or arm, and if the distro is debian or ubuntu, since these options are now available prebuilt by beamMP. I will look into creating that

Give me 5 min, I do it for you.

O1LER:
Ive done something wrong when updating my main branch i guess, i dont want to revert my changes from May

QuintenQVD0:
> Ive done something wrong when updating my main branch i guess, i dont want to revert my changes from May

That's fine, I handle that for you.

QuintenQVD0 (approved):
LGTM! Thanks

QuintenQVD0:
Thank you for bringing this to the repo! Now the egg will based on the host arch download the debian 11 arm64 one or the debian 11 x86_64 one

O1LER:
Ok, nice. I guess you left the last check for Server-linux in place for backwards compatibility should someone choose an older version before v3.2.2

QuintenQVD0:
> Ok, nice. I guess you left the last check for Server-linux in place for backwards compatibility should someone choose an older version before v3.2.2

I could do that. let me quickly implement that

O1LER:
I mean it is still there, its in the function that prints `No valid versions found. Defaulting to the latest release`

QuintenQVD0:
There is no real need for supporting older version as they only support the latest version.
I modified the fallback logic so that it tries to find the x64 version with the keywords debian and server

QuintenQVD0:
> I mean it is still there, its in the function that prints `No valid versions found. Defaulting to the latest release` that then goes to grab `Server-linux` but this binary will not be published anymore in the future

The server-linux does not run in pterodactyl anymore.

O1LER:
> > I mean it is still there, its in the function that prints `No valid versions found. Defaulting to the latest release` that then goes to grab `Server-linux` but this binary will not be published anymore in the future
> 
> The server-linux does not run in pterodactyl anymore.

Since that was the ubuntu version i guess?

Thanks for the help, much appreciated

PR #2670, opened 2024-01-19
Update Vintage Story egg

Th3Dilli:
# Description

This PR fixes issues with installing specific version of vintagestory.

In version 1.18.8 the game was updated to dotnet 7 and the urls have changed, since it is now platform depended. This was partially fixed already with the last update but older version where using the old urls so you could only install pre dotnet 7 version but those wouldn't work since the new egg only has the dotnet image to choose from.

Now a users will be only able to download versions since 1.18.8 which are going to work with the dotnet 7 yolks image this egg uses.

Further we need to clean up the assets folder since old installation files that are not included in a new release may cause issues when loaded by the game. We also now delete the Lib folder to clean up old unused libraries as well. The other files and Mods folder are fine since they always contain the same files so those will be overwritten.

And the last thing this PR addresses is that the new 1.19 version includes a "/db vacuum" command that clear the sqlite db file from any unused rows and reduces it size.
The issue we ran into there was that it by default sqlite uses the /tmp directory to create the new db file but a usual vs savegame (.vcdbs) is larger then 100M as the default /tmp directory in wings and so it would always fail with "database or disk full" error and crash.
So now we redirect the location for the tmp file to the /home/container/data/Backup folder which will exist once the game has started.
A user will not need to change that directory. And hosters can still change that if they changed the data directory for the startup command.


QuintenQVD0:
- I would like to keep the fallback logic if the branch is not found.
- Please keep `  echo "API URL: $API_URL"`
- then all this logic is not needed, if it fails, just fallback to the latest stable one
```bash
if [ $? -ne 0 ]; then
  echo "-----------------------------------------"
  echo "Installation failed..."
  echo "Please make sure the specified version exists: $RELEASE_VERSION"
  echo "-----------------------------------------"
  ```

Th3Dilli:
I thought it would be a better user experience this way.

For instance, if a user selects a branch and enters a version other than "latest," the system will now return an error if the specified version cannot be found (due to a potential mistype or removal from our server), rather than defaulting to another version. 
It also may cause issues when the server automatically updates to a different version than what was input. (Mods most likely need to be updated before starting the server to ensure they work on the current version)

Eventually, the output could include that the user should switch to RELEASE_VERSION="latest" if that's what they wanted.


The Echo of the API URL remains available in scenarios when choosing a branch (stable/unstable) and leaving "latest" in the RELEASE_VERSION field, the system will query the respective API endpoint to fetch the latest version. 
The obsolete pre API endpoint has been removed, and the logic has been adjusted to disallow its use with "latest."


The previous fallback logic, which was positioned after the completion of operations related to the API endpoint did only fall back to stable to then echo the API endpoint which did not functionally change anything. Further, the  FILES_BRANCH can only be set via dropdown so there is/was no way to set it to an invalid branch as far as I can tell.

Now the check with `if [ $? -ne 0 ]; then` will make sure if the user specified a version via RELEASE_VERSION and it did not exist or was removed it does so tell them.

QuintenQVD0 (approved):
LGTM! thanks

PR #2697, opened 2024-01-27
feat(generic): expose python 3.12

jayceslesar:
# Description

Looks like you already made a upstream change to support 3.12 so I am just exposing it here.

Would be really cool to figure out how we can support https://github.com/pdm-project/pdm and https://python-poetry.org/ to install python packages through them.

Cheers


parkervcp (approved):

parkervcp:
I would have preferred the term "add" to "expose".

QuintenQVD0:
And that this egg was fully exported from the panel..

But thank you for the change

PR #2706, opened 2024-01-30
Update Palworld

QuintenQVD0:
# Description

- Add and download a self written parser application for their (bad) config file
- Add a working console over RCON
- allow toggling bEnableInvaderEnemy what is known for memory leaking
 
 If anyone want to see the source of the parser application: https://github.com/QuintenQVD0/Palword-server-config-parser and yes it does also work on Windows even tho that is not pterodactyl related.
 
 **For Those updating an exiting egg / server**
 1. Update your egg
 2. Update the startup of all already made servers to the one that now comes with the egg
 3. Then hit reinstall because the parser application has to be downloaded 

**For those testing this egg**
This will not work as long as this is not merged
If you want to test this egg, on line 64 replace the link:
-> `https://raw.githubusercontent.com/parkervcp/eggs/master/game_eggs/steamcmd_servers/palworld/PalworldServerConfigParser-linux-amd64` 
to:
-> `https://github.com/QuintenQVD0/Palword-server-config-parser/releases/latest/download/PalworldServerConfigParser-linux-amd64`

  * If not, why?:


GameAreaCZ:
Works fine! Thank you for your work.

lilkingjr1:
How is the parser application used? It is a welcomed addition, but the install script only downloads it and it doesn't seem to be used again.

QuintenQVD0:
> How is the parser application used? It is a welcomed addition, but the install script only downloads it and it doesn't seem to be used again.

Its in the startup. Its the first things that runs.

Toetje585:
Works awesome, thanks for the effort!

lilkingjr1:
@QuintenQVD0 Oh my goodness, my bad, totally missed that.

lilkingjr1:
@QuintenQVD0 If I set a setting (like `DeathPenalty=None`) in the config that is not currently handled by your script, will it get overwritten?

parkervcp (approved):
This is working so far as I am concerned.

j-barot:
I don't know how to use RCON ... Anyone can explain me ? ^^

QuintenQVD0:
> @QuintenQVD0 If I set a setting (like `DeathPenalty=None`) in the config that is not currently handled by your script, will it get overwritten?

No. The application will not touch variables I have not set to be parsed. So as in your example if you manual edit the DeadPenalty to something else then that will just stay as you set it. Until you hit reinstall because then a fresh config is made.

PR #2773, opened 2024-02-27
[Update] egg-space-engineers.json default startup paramters

Dr3nz4r:
so that the server outputs into the console and can be detected as running

# Description

Update the existing startup parameters to ensure that the server logs into the console. This is required for the server so it can detect online status and print logs into the ptero console. 
Before the change the server would be kept in "starting" and only output this into the console (even though the server was booted and writing logs to file and allowing players etc.)

```
:/home/container$ export WINEDLLOVERRIDES="mscoree=n,b;mshtml=n,b"; wine /home/container/DedicatedServer64/SpaceEngineersDedicated.exe -path Z:\\home\\container\\config -noconsole -ignorelastsession
```

  * If not, why?:

  * didn't do this because it's way more effort than the change itself. Especially as it's a super simple one


QuintenQVD0:
I appended some needed updates like that users can not change ports (as that was never allowed), moved to vcrun2022 (includes vcrun2010-2022) and hard locked the app ID

QuintenQVD0 (approved):
LGTM! thanks you

PR #2787, opened 2024-03-07
Update egg-mongo-d-b7.json

Turkidev:
replacing the 127.0.0.1 with {{SERVER_IP}} to fix when the node has more than 1 ip

# Description

  * If not, why?:

  * [X] If some script is needed then it is part of a current yolk or a PR to add one

## New egg Submissions

3. [ ] Have you added the egg to the main README.md and any other README files in subdirectories of the egg (e.g /game_eggs) according to the alphabetical order?
4. [ ] Have you added a unique README.md for the egg you are adding according to the alphabetical order?
5. [ ] You verify that the start command applied does not use a shell script
6. [ ] The egg was exported from the panel

### DIFF
--- a/database/nosql/mongodb/egg-mongo-d-b7.json
+++ b/database/nosql/mongodb/egg-mongo-d-b7.json
@@ -13,7 +13,7 @@
         "MongoDB_7": "ghcr.io\/parkervcp\/yolks:mongodb_7"
     },
     "file_denylist": [],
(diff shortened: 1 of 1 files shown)


Turkidev:
> then you must change the image too, because it runs also on 127.0.0.1. I think, this change is not needed, also when the server have multiple IP's

no need to change the image just change the startup command and it works fine now

QuintenQVD0:
This is not needed in any way as it binds to 0.0.0.0 what means all interfaces, if you have issues then enable on the egg the Force Outgoing IP

PR #2810, opened 2024-03-19
Update Valheim BepINex API url

thomasbonsirven:
# Description

the sub-domaine not work for install BepInEx mod.
Error 404.

Without sub-domain is work.


QuintenQVD0:
this was not exported from the panel and you litterly removed the whole template.

At least re-export this from a panel and fully replace it here, else this will be converted to a draft

thomasbonsirven:
Aaah, ok.
I don't have to modify the egg directly in GitHub, but in the panel and then export it to copy/paste it into gihub every line. Is that correct?

QuintenQVD0:
> Aaah, ok. I don't have to modify the egg directly in GitHub, but in the panel and then export it to copy/paste it into gihub every line. Is that correct?

you just import the egg, make the needed to be changed in the panel, export it. Then replace the full file here with the one you got from your export

ImKringle:
Check other Update PRs for how this is done, you replace the updated lines not the entire egg

QuintenQVD0 (approved):
LGTM!

thomasbonsirven:
Hey,

Thanks you @QuintenQVD0, for the export :)
I dont have the time this morning.

PR #2811, opened 2024-03-19
Update power-nukkitx 2.0

JblusItsMe:


QuintenQVD0:
Their is more things wrong with this pr

This is not an update but a whole new egg

You literly removed the whole template

ImKringle:
This is again just a fully new egg

If making an update, only replace the lines **you** changed; Adding an entirely new egg is only if its getting its own section in the repo. This is an update, so only replace what is required. A whole new egg isnt required

Re:
`Their is more things wrong with this pr

This is not an update but a whole new egg

You literly removed the whole template`

ImKringle (changes requested):
Replace just whats changed. Install script, variables, etc - not upload a whole new file

PR #2818, opened 2024-03-23
Palworld: Added instructions for changing config

Maelstromeous:
# Description

It was unclear as to why config changes were not being applied. After hunting the file path down in the **source code of the generator** this readme required an update.

* [x] N/A Have you tested and reviewed your changes with confidence that everything works?


QuintenQVD0:
Ofc it does not the name sas it DEFAULT.

And on every startup the parser will print the full path to the terminal to where the config it is writing to it?

Maelstromeous:
Yes but you shouldn't have to dig into the generator **or the server logs** to figure out where the config file path is, when it is much simpler to state in the readme.

QuintenQVD0:
> Yes but you shouldn't have to dig into the generator **or the server logs** to figure out where the config file path is, when it is much simpler to state in the readme.

I merge it that is not the issue, that fact is just that every  time you hit start the console will tell you at the top where the config is.

afbeelding

QuintenQVD0 (approved):

PR #2826, opened 2024-03-25
fix Enshrouded console output

QuintenQVD0:
# Description

- fixes the no console output because of proton and wine updates

  * If not, why?:


parkervcp (approved):

ValhallaCommunityOfficial:
Fresh install and I cant get it to show on server browser. Shows when i search via ip but when adding to favorites, does not show on top of server list. No way to join

ValhallaCommunityOfficial:
Yeah, starting to realize that now. thanks for the reply

PR #2880, opened 2024-04-09
Update arma reforger

### LINKED ISSUE #2879: [Bug]: Arma Reforger curl timeout (fix found and validated)
### Panel Version

1.1.15

### Wings Version

1.1.11

### Service

Arma Reforger

### Modified

No, I did not modify the egg

### Expected Behavior

Install server, add mods on the config file restart server starts

### Actual Behavior

Same problem as in https://github.com/parkervcp/eggs/issues/2783
You closed it but problem lies with arma workshop needing a /tmp folder configured.
This has been verified by Arma Reforger developers

image
-addonTempDir with a directory needs to be added to launch parameters and it will fix the problem.

Has been confirmed by all my clients.

### Steps To Reproduce

Install server
Add mod in config
error

add -addonTempDir
restart server
mod downloads fine

### Install logs

https://pteropaste.com/6ngpo2n8ew6i/

QuintenQVD0:
# Description

- Set a default addonTempDir
- closes #2879 

  * If not, why?:


lilkingjr1:
https://github.com/parkervcp/eggs/issues/2879 implies only the startup flag needs to be added to resolve the issue. If the Arma server can automatically create the missing folder, I don't think we need to adjust the install script. Also, I don't think we need a variable for the folder location. I would just hardcode it into the startup command personally (that way an end-user doesn't accidentally put it in a reserved folder that somehow breaks things).

lilkingjr1:
> a company may want it configurable to match mounts they have configured

But it's a temp/throwaway folder just for downloading, no? Why would you assign it to a mount? Only reason I can think of would be to give free space other than what is allocated.

parkervcp:
> > a company may want it configurable to match mounts they have configured
> 
> Why would you assign it to a mount? Only reason I can think of would be to give free space other than what is allocated.

I can think of many reasons. They could mount to hdd based storage instead of ssd based to save on writes to the disk and have the disks last longer.

lilkingjr1:
That's fair. I also just realized the variable is not user-editable, which works for me 👍

PR #2869, opened 2024-04-06
Fix Palworld - Added SteamSDK to Install

ImKringle:
# Description

Closes #2866 - PocketPairs SH file is supposed to do this on its own. Not using it requires its added in the install

Relevant test that it fixed the issue: https://github.com/parkervcp/eggs/issues/2866#issuecomment-2041199579
## Checklist for all submissions

* [x] Have you followed the guidelines in our [Contributing document](https://github.com/parkervcp/eggs/blob/master/CONTRIBUTING.md)?
* [x] Have you checked to ensure there aren't other open [Pull Requests](../pulls) for the same update/change?
* [x] Have you tested and reviewed your changes with confidence that everything works?
* [x] Did you branch your changes and PR from that branch and not from your master branch?
  * If not, why?:

* [x] You verify that the start command applied does not use a shell script
  * [ ] If some script is needed then it is part of a current yolk or a PR to add one
* [x] The egg was exported from the panel

PR #2883, opened 2024-04-10
Fix - The Forest

ImKringle:
# Description

As of now the File parse type reads both the serverPassword and serverPasswordAdmin section of the config.cfg as just the main server password. Doing this causes a crash on startup as two sections of the config are the same thing

This edit changes it to not include the serverPasswordAdmin in the config and instead sets it in the Startup Command

This edit is for:
1. Removes serverPasswordAdmin from config creation on install
2. Adds -serverPasswordAdmin {{ADMIN_PASS}} to startup
3. Removes serverPasswordAdmin from the configuration files section of the egg

  * If not, why?:


ImKringle:
erm.. okay this for some reason includes my PR for #2869 - Lemme fix that >.>

PR #2923, opened 2024-04-30
Remove ` --disable-ansi` flag from PMMP egg

SmallkingDev:
# Description

PocketMine-MP can detect whether the system has support for using ansi. There is no need to leave this flag at startup by default.


QuintenQVD0:
This is to the wrong repo, please move it this PR to the minecraft repo of the pelican organisation

SmallkingDev:
> This is to the wrong repo, please move it this PR to the minecraft repo of the pelican organisation

Ok

PR #2929, opened 2024-05-05
Added Java Version 11 to nano limbo egg

ShortByte:
Added new java version (Java version 11) to the Nano Limbo Egg for Minecraft.
This Java version is required by the newest version of Nano Limbo.


QuintenQVD0:
now export if from a panel and fully replace it and then to the right repo (this one is not used anymore, use the minecraft one on this organisation)

PR #2930, opened 2024-05-05
Update egg-paper.json

Oldbear200:
added java 19 because i needed it for smth so eventually do you


gOOvER:
Jave 19 is not any LTS version. And you are the only, who need it. So no need to add this

QuintenQVD0:
yea java 19 is not needed for any specific mc version so I am closing this.

